PR #22, opened 2022-05-20
Eliminate easy warnings, fix some code issues.

byrnHDF:
Fixed the easy warnings for unused variables/parameters and array initialization.

Added missing xdr structure member - removed some code that may need to be reverted after testing.

WIP


hdf/examples/AN_create_annotation.c
```diff
@@ -1,5 +1,11 @@
 #include "hdf.h"
 
+/* Used to make certain a return value _is_not_ a value.  If not ture, */
```
lrknox, line 3: Is "ture" a misspelling of true?  Also, it looks like CHECK(ret, val, where) is checking to see that the return value is not the same as value val.  It could be some value as long as it was a different value?  The comment could be interpreted that the CHECK was to make certain ret was not any value.

hdf/examples/AN_get_annotation_info.c
```diff
@@ -1,5 +1,11 @@
 #include "hdf.h"
 
+/* Used to make certain a return value _is_not_ a value.  If not ture, */
```
lrknox, line 3: Ditto

hdf/examples/GR_create_and_write_image.c
```diff
@@ -1,12 +1,18 @@
 #include "hdf.h"
 
+/* Used to make certain a return value _is_not_ a value.  If not ture, */
```
lrknox, line 3: Ditto

hdf/examples/GR_get_attribute.c
```diff
@@ -1,9 +1,15 @@
 #include "hdf.h"
 
+/* Used to make certain a return value _is_not_ a value.  If not ture, */
```
lrknox, line 3: Ditto

hdf/examples/GR_image_info.c
```diff
@@ -1,8 +1,14 @@
 #include "hdf.h"
 
+/* Used to make certain a return value _is_not_ a value.  If not ture, */
```
lrknox, line 3: Ditto

hdf/examples/GR_modify_image.c
```diff
@@ -1,9 +1,15 @@
 #include "hdf.h"
 
+/* Used to make certain a return value _is_not_ a value.  If not ture, */
```
lrknox, line 3: Ditto

hdf/examples/GR_read_chunks.c
```diff
@@ -1,55 +1,60 @@
...
-   /************************* Variable declaration **************************/ 
-
-   intn  status;         /* status for functions returning an intn */ 
-   int32 file_id,        /* HDF file identifier */ 
-         gr_id,          /* GR interface identifier */ 
-         ri_id,          /* raster image identifier */ 
-         dims[2],        /* dimension sizes of the image array */ 
+#include "hdf.h"
+
+/* Used to make certain a return value _is_not_ a value.  If not ture, */
```
lrknox, line 3: Ditto

hdf/examples/AN_create_annotation.c
byrnHDF: I wasn't looking at comments - just warnings. We should a spellchecker if we want perfect spelling.

byrnHDF:
I suggest a spellcheck as a separate PR.

lrknox:
That's fine, I suppose it will catch all occurrences of "ture".  It won't check for the meaning of the comment.

________________________________

hdf/examples/AN_create_annotation.c
bmribler: Ha..ha...  44 of them!  I think it should be "If it is,".  I'll fix them after you merge and I check out to go through the CHECK/VERIFY macros as we agreed before.

hdf/examples/AN_get_annotation_info.c
```diff
@@ -1,5 +1,11 @@
...
+/* print error messages, increment num_err and return. */
+#define CHECK(ret, val, where) \
+do {if(ret == val) {printf("*** ERROR from %s is %ld at line %4d in %s\n", where, (long)ret, (int)__LINE__,__FILE__);} \
+} while(0)
+
```
bmribler, line 8: A header file should be added and this macro should go in there instead of every .c file.  On the other hand, that structure could be an issue for the examples to be posted on the portal pages... Could it?  I looked in hdf5 to see how it was done.  The returned values are not checked.  I think years ago, we agreed on that for examples...

byrnHDF: That was the first idea, but there was no header file, second best add the macro.
One more option - add the macro to a public header that does get included.

bmribler: Why can't we just add a new header file for the examples, like in the test directory?  Will that cause complication that I don't know?

byrnHDF: Another header then has to be managed - does it get published (handled by make install) , is it safe and documented for use by users, are there dependencies. Should we include other useful macros. Maybe create a test macros header, that just included by another header. All this for a very legacy product?

bmribler: I see your points, but I don't think it will be complicated because it will be local to the examples directory only.  Managing it is as simple as managing hdftest.h in the test directory.  Also, you might run into the duplicate definition of CHECK issue if the examples interact with each other somehow in the future.  It's very unlikely but not impossible.  Regarding your concerns:
- does it get published (handled by make install): yes, for users to use it too
- is it safe and documented for use by users: yes, it will be a very simple header file
- are there dependencies: no reasons to be
- Should we include other useful macros: definitely for these examples, that's the essential benefit of its existance

Sorry, you lost me here: Maybe create a test macros header, that just included by another header.

Another significant and much likely benefit is avoiding an error-prone situation.  If you want to change something in the macros, only one version will be changed, instead of many source files, as you already saw for correcting "If not ture,".

I'm willing to do the work if that will help.

byrnHDF: I agree, one file to include is best.
It's a small useful macro - just add it to an appropriate public header that already is installed.

bmribler: OK.  At the end of hdf.h is good since both sd and hdf examples can have it.  Can you reverse the addition of the macro definition to the source files?  I'll add it after you merge.  If it's too much hassle to reverse, I can manually remove them.

hdf/src/hdf.h
```diff
@@ -145,6 +145,17 @@ typedef intn (*hdf_termfunc_t)(void);   /* termination function typedef */
 
 /* .................................................................. */
 
+/* Publically useful macros for tests/examples.  */
+
+/* Used to make certain a checked value _is_not_ a specific value.
```
lrknox, line 150: if(ret == val) an error message is printed.  I still think the comment should read "If true, meaning that th checked value matches the specific value, print an error message."

mfhdf/hdfimport/gen_sds_floats.c
```diff
@@ -13,59 +13,50 @@
...
+#define SDS1_NAME      "SDStemplate"
+#define SDS2_NAME      "SDStemplate again"
 #define X_LENGTH      6
 #define Y_LENGTH      9
 #define Z_LENGTH      2
 #define RANK2          2  /* Number of dimensions of the SDS */
 #define RANK3          3  /* Number of dimensions of the SDS */
 
-main() 
+int main()
```
lrknox, line 26: Won't this main function need to return an int now?

PR #28, opened 2022-11-12
Fix build errors on Windows for ARM64

cgohlke:
Using Visual Studio 2002 17.4 on Windows 11 for ARM64.


derobins:
We are currently reworking the build system so this may not be necessary. I'm going to leave this open for reference as we work and we'll close / merge as needed.

PR #40, opened 2022-12-26
Remove most blocks disabled by `OLD_WAY` if defineds.

schwehr:


byrnHDF (approved):

hdf/test/litend.c
```diff
@@ -309,50 +309,6 @@ test_little_read(void)
...
-            for(j=0; j<sizeof(float64); j++)
-              printf("%.2x ",u8_d2[j]);
-            printf("\n");
-}
-              HEprint(stdout,0);
-              num_errs++;
-            } /* end if */
-	  } /* end for */
-}
-#endif /* OLD_WAY */
```
bmribler, line 355: Is there a reason to remove both #if and #else blocks?

derobins: This needs to be fixed before merging. We're probably going to yank all the "old machine" stuff (the convex float bits, so this is probably dead code), but let's do that comprehensively in a separate PR.

derobins (changes requested):
Need to fix the one, over-enthusiastic edit

hdf/test/litend.c
schwehr: Reverted the whole file and then pushed a change that is hopefully just removing the old code.  Thank you for catching the mistake!

derobins (approved):

derobins:
Perfect. Let's get rid of this cruft.

PR #43, opened 2022-12-27
Use double

schwehr:


schwehr:
This pr is way out of date and should happen after the next release.  Closing.

PR #44, opened 2022-12-27
Use float literals instead of casts: (float) 1.2 -> 1.2f

schwehr:


schwehr:
Doing this wasn't hard, there has a been a ton of churn, and the change should probably happen after the next point release.  Closing the PR.  If you want me to catch it up and do this, please reopen the pr.

PR #45, opened 2022-12-27
Revert ubuntu version to 20.04 from latest.

lrknox:
HDF4 can't find xdr functions in ubuntu-22.04 as now defined for ubuntu-latest.  This change to ubuntu-20.04 will allow tests to pass for now;  the problem with ubuntu 22.04 should be investigated and addressed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
 #            generator: "-G \"Visual Studio 16 2019\" -A x64"
           - name: "Ubuntu Latest GCC"
             artifact: "Linux.tar.xz"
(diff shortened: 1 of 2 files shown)


schwehr:
Probably something related to https://packages.ubuntu.com/jammy/libtirpc-dev

I work in a different env, but had issues as xdr was removed from the core of libc and had to add an explicit dependency on libtirpc-dev

bmribler (approved):

lrknox:
Thank you for the suggestion!

I’m going to merge the reversion to ubuntu 22.04 for now, and will try out installing libtirpc-dev in a few days after we release HDF5 1.14.0.

PR #46, opened 2022-12-29
mfhdf/test/tdatainfo.c: General cleanup

schwehr:
Demonstrate initial clean up of one test file

- include what you use (iwyu) but then some hand editing
- remove signatures for comments
- remove definition of DATAINFO_TESTER
- remove redundant test function prototypes
- convert `#define`s to const int and const char *
- localize variables / combine definition and initialization
- add const
- use stdint.h: int16 and int32 -> int16_t and int32_t
- intn -> int
- add static on more local functions
- remove empty statement: `;`
- remove `#if 0` block of code
- remove commented out checks

### DIFF
--- a/mfhdf/test/tdatainfo.c
+++ b/mfhdf/test/tdatainfo.c
@@ -16,80 +16,72 @@
  *    test_datainfo - test driver
  *  test_nonspecial_SDSs - tests nonspecial SDSs
  *  test_compressed_SDSs - tests compressed SDSs without closing file
- *  test_empty_SDSs      - tests on empty chunked and chunked/comp SDSs
+ *  test_chunked_empty_SDSs      - tests on empty chunked and chunked/comp SDSs
  *  test_chunked_partial - tests on chunked and partially written SDS
  *  test_chkcmp_SDSs     - tests chunked/compressed SDSs
  *  test_extend_SDSs     - tests SDSs with unlimited dimensions
- * -BMR, Jul 2010
  ****************************************************************************/
 
 #include "mfhdf.h"
 
-#ifdef H4_HAVE_SYS_TYPES_H
-#include <sys/types.h>
-#endif
-#ifdef H4_HAVE_SYS_STAT_H
-#include <sys/stat.h>
-#endif
+#include <math.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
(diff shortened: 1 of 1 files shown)


mfhdf/test/tdatainfo.c
```diff
@@ -943,113 +777,102 @@ static intn test_compressed_SDSs()
  * number of blocks for data, which should be 0.
  */
 /****************************************************************************
- Name: test_empty_SDSs() - tests special but empty SDSs
+ Name: test_chunked_empty_SDSs() - tests special but empty SDSs
```
bmribler, line 780: The term "chunked" is incorrect in this test function.

mfhdf/test/tdatainfo.c
```diff
@@ -1207,35 +1003,21 @@ static intn test_chunked_partial()
     /* Record number of values the SDS can have */
     sds_info.n_values = 1 * 10; /* chunk has 1 dim of size 10 */
 
-#if 0 /* need to figure out how to test this feature */
-    status = SDgetdatainfo(sds_id, NULL, 0, info_count, sds_info.offsets, sds_info.lengths);
-    CHECK(status, FAIL, "test_chunked_partial: SDgetdatainfo");
-#endif
-
```
bmribler, line 1214: This feature should be tested instead of removed.

mfhdf/test/tdatainfo.c
```diff
@@ -1755,28 +1505,14 @@ static intn test_extend_SDSs()
...
     TESTING("getting location info of data (tdatainfo.c)");
 
-    /* Test nonspecial SDSs */
-    num_errs = num_errs + test_nonspecial_SDSs();
-
-    /* Test compressed SDSs */
+    int num_errs = test_nonspecial_SDSs();
     num_errs = num_errs + test_compressed_SDSs();
-
-    /* Test chunked empty SDSs */
```
bmribler, line 1770: Comment from the original code was incorrect.  Should be "Test special empty SDSs".

schwehr:
Thanks for the feedback.  I'm probably not the right person to be doing the fixups an tdatainfo.c.  Also a lot of cleanup has already happened.  I'm going to close this PR.  Feel free to reopen if you'd like me to give this file a go.

bmribler:
OK, we will see what are left here after all the cleanups are done.  Thank you!!

PR #49, opened 2023-01-04
Cleanup glist

schwehr:
This assumes that users are not counting on the rest of glist outside of hdf4.  It would also possible to `#if` wrap all the unused code and have users set that define to get all the extraneous functions.

- Remove almost all unused code
- IWYU
- Remove unnecessary comments
- intn -> int
- Localize variables / combine definition and initialization
- Remove gotos
- Just return a value as soon as possible

Only use in hdf4:

```
find . -type f | egrep -v '[.]git$|build-ninja| |glist' | xargs grep HDGL
./hdf/src/hfile.c:          if (HDGLinitialize_list(cleanup_list) == FAIL)
./hdf/src/hfile.c:    if(HDGLadd_to_list(*cleanup_list,(void *)term_func)==FAIL)
./hdf/src/hfile.c:    if((term_func=(hdf_termfunc_t)HDGLfirst_in_list(*cleanup_list))!=NULL)
./hdf/src/hfile.c:          } while((term_func=(hdf_termfunc_t)HDGLnext_in_list(*cleanup_list))!=NULL);
./hdf/src/hfile.c:    HDGLdestroy_list(cleanup_list);    /* clear the list of interface cleanup routines */
```


schwehr:
The point of this pull request is mostly just for discussion. It's now way out of date and there have been lots of discussions elsewhere on this topic along with lots of cleanup CLs, so I'm closing this pr.  Especially with it being way out of date and needing a merge.

PR #50, opened 2023-01-05
Porting to m1 - part 1 (initial attempt to pass tests on m1)

bmribler:
Description:
    - Removed some outdated machines from #ifdef lines as a temporary approach
      until autoconf approach works.
    - Added __APPLE__ to some of those lines, also as temporary approach
    - Replaced "free()" calls by "HDfreenclear()" to reset pointers after freeing
    - Removed implicit declaration of function errors, some need attention
    - Corrected string comparison in several if statements
    - Fixed many typos
Platforms Tested:
    Darwin (m1)
    Linux/64 (jelly)
    Note that at this time, hdf and mfhdf tests passed.  ncgen test still failed
        dues to one data value being corrupted.  It is hopeful that the autoconf
        approach will address that issue.

### DIFF
--- a/hdf/src/hconv.h
+++ b/hdf/src/hconv.h
@@ -59,7 +59,7 @@
 /* CONSTANT DEFINITIONS                                                      */
 /*****************************************************************************/
 /* Generally Big-Endian machines */
-#if !defined(INTEL86) && !defined(MIPSEL) && !defined(DEC_ALPHA) && !defined(I860) && !defined(SUN386) && !(defined(__ia64) && !(defined(hpux) || defined(__hpux))) && !defined(__x86_64__)
+#if !defined(__APPLE__) && !defined(INTEL86) && !defined(MIPSEL) && !defined(I860) && !defined(SUN386) && !(defined(__ia64) && !(defined(hpux) || defined(__hpux))) && !defined(__x86_64__)
 #       define UI8_IN     DFKnb1b   /* Unsigned Integer, 8 bits */
 #       define UI8_OUT    DFKnb1b
 #       define SI16_IN    DFKnb2b   /* S = Signed */
@@ -90,7 +90,7 @@
 #       define LF64_IN    DFKsb8b
 #       define LF64_OUT   DFKsb8b
 
-#else  /* must be INTEL86 || MIPSEL || DEC_ALPHA || I860 || SUN386 || IA64 || Linux64 (Generally, little-endian machines */
+#else  /* must be __APPLE__ || INTEL86 || MIPSEL || I860 || SUN386 || IA64 || Linux64 (Generally, little-endian machines */
 #   define UI8_IN     DFKnb1b   /* Big-Endian IEEE support */
 #   define UI8_OUT    DFKnb1b   /* The s in DFKsb2b is for swap */
 #   define SI16_IN    DFKsb2b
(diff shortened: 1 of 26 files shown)


hdf/src/hdatainfo.h
```diff
@@ -11,7 +11,7 @@
  * help@hdfgroup.org.                                                        *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
-/* $Id: hproto.h 5400 2010-04-22 03:45:32Z bmribler $ */
+/* $Id: hdatainfo.h 5400 2010-04-22 03:45:32Z bmribler $ */
```
byrnHDF, line 14: Is this line necessary?

derobins: No. These are an artifact of CVS.

bmribler: I know, the other PR took care of that, so I just left it there for github to do its thing.  I can remove now if you think I should.

bmribler:
One of the committed files was wrong.

hdf/src/hdatainfo.h
bmribler: Allen, do you receive this notification?

hdf/src/hconv.h
```diff
@@ -59,7 +59,7 @@
 /* CONSTANT DEFINITIONS                                                      */
 /*****************************************************************************/
 /* Generally Big-Endian machines */
-#if !defined(INTEL86) && !defined(MIPSEL) && !defined(DEC_ALPHA) && !defined(I860) && !defined(SUN386) && !(defined(__ia64) && !(defined(hpux) || defined(__hpux))) && !defined(__x86_64__)
+#if !defined(__APPLE__) && !defined(INTEL86) && !defined(MIPSEL) && !defined(I860) && !defined(SUN386) && !(defined(__ia64) && !(defined(hpux) || defined(__hpux))) && !defined(__x86_64__)
```
derobins, line 62: Ideally this will all disappear. Feature tests will replace this as much as possible.

hdf/src/hdfi.h
```diff
@@ -88,6 +88,10 @@
 #include <limits.h>
 #include <string.h>
 
+/* Temporarily eliminate implicit declaration of funcion 'yyerror' */
+int yylex(void);
+void yyerror(char* s);
```
derobins, line 93: It looks like this is from the parser generator? Usually we can stick this in the generator script.

derobins: This can go in a %top directive in the source .y/.l file(s)

mfhdf/hdiff/hdiff.h
```diff
@@ -4,13 +4,10 @@
...
  *********************************************************************/
 
-
-
-
 #ifndef HDIFF_H__
 #define HDIFF_H__
 
-
+#include <unistd.h>
```
derobins, line 10: I feel like this should go in the main header file. This will also fail on Windows, so it needs to be ifdef-protected.

mfhdf/hdiff/hdiff.h
```diff
@@ -193,6 +190,13 @@ uint32 diff_sds(int32 sd1_id,
                 int32 ref2,
                 diff_opt_t *opt);
 
+/* added to temporarily remove implicit declaration of func until can fix better */
+ /* extern int      getopt          PROTO((
+                                       int  argc,
+                                       char **argv,
+                                       char *opts
+                                       ));
+ */
```
derobins, line 199: This should not be necessary.

mfhdf/libsrc/xdrposix.c
```diff
@@ -54,11 +54,7 @@ typedef NETLONG     netlong;
 /*EIP #include "netcdf.h" */
 #include "mfhdf.h"
 
-#if !(defined DOS_FS)
```
derobins, line 57: lol

PR #56, opened 2023-01-07
Removes the MANIFEST file and updates the release scripts

derobins:


move-if-change
```diff
@@ -1,15 +0,0 @@
-#!/bin/sh
```
derobins, line 1: I have no idea why this script is needed. It's not used anywhere in the HDF4 source.

bin/release
```diff
@@ -20,11 +20,10 @@
 USAGE()
 {
 cat << EOF
-Usage: $0 -d <dir> [-h] [--nocheck] [--private] <methods> ...
```
derobins, line 23: I made this file work like HDF5's release script

bin/bbrelease
```diff
@@ -14,64 +14,56 @@
 
 # Make a release of hdf4.
 #
-# Programmer: Robb Matzke
```
derobins, line 17: I made this file work like HDF5's bbrelease script

lrknox (approved):

PR #58, opened 2023-01-08
Tidies configure.ac

derobins:
* Adds silent rules so we can finally spot warnings
* Removes a couple of obsolete things from configure.ac
* Adds HDF5's maintenance mode scheme and the switch script
* Quiets autoreconf warnings
* Ran autoupdate to update some macros for better cross-compiling


configure.ac
```diff
@@ -750,8 +768,6 @@ fi
 
 AM_CONDITIONAL([HDF_BUILD_XDR], [test "X$BUILD_XDR" = "Xyes"])
 
-AC_HEADER_STDC
-AC_HEADER_RESOLV
```
derobins, line 754: These are both obsolete and resolv.h is not used in the HDF4 source

configure.ac
```diff
@@ -613,7 +633,7 @@ if test "X$HAVE_SZIP" = "Xyes" -a "x$HAVE_SZLIB_H" = "xyes"; then
 
     AC_SUBST([LL_PATH]) LL_PATH="$LD_LIBRARY_PATH"
 
-    AC_TRY_RUN([
+    AC_RUN_IFELSE([AC_LANG_SOURCE([[
```
derobins, line 636: The macro changes in the rest of the file were upgraded by autoupdate to be nicer for cross-compiling

byrnHDF (approved):

PR #59, opened 2023-01-08
Fixes misc warnings, mainly format string problems

derobins:


derobins:
Most of the warnings I fixed were issues in the tools where %ld was specified for an int32.

byrnHDF (approved):

PR #62, opened 2023-01-09
Removed warning from -Woverlength-strings

bmribler:
Description:
    The usage was stored in a long string, causing this warning:
    warning: string literal of length 708 exceeds maximum length 509 that C90 compilers are required to support [-Woverlength-strings]

    Changed to use printf per line instead.  If this causes issue with
    ncdump user scripts then it'll have to be reverted.


derobins (approved):

byrnHDF (commented):
okay

PR #65, opened 2023-01-09
Removed implicit declaration warning of HDiscdf/HDisnetcdf/HDisnetcdf64

bmribler:
Description:
    Moved the declaration of HDiscdf, HDisnetcdf, HDisnetcdf64 from
    a private header to a public one.  They are used in tests and
    can be used by user applications.

### DIFF
--- a/mfhdf/libsrc/local_nc.h
+++ b/mfhdf/libsrc/local_nc.h
@@ -825,15 +825,6 @@ HDFLIBAPI bool_t nssdc_write_cdf
 HDFLIBAPI bool_t nssdc_xdr_cdf
     PROTO((XDR *xdrs, NC **handlep));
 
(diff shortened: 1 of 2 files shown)


derobins (approved):

derobins:
Wait, shouldn't this have a release note of some sort if we're adding it to the public API? Also, is it a good idea to expose internal convenience functions via the public API?

derobins (changes requested):
Probably needs a release note, at least.

bmribler:
> Probably needs a release note, at least.

I'll do that tomorrow.  Although, I kind of having second thought.  Please don't merge this one yet.

lrknox (approved):

bmribler:
Decision: will not do this change at this time because it's only warning now.

PR #68, opened 2023-01-09
Hdf4 bmr typos

bmribler:


mfhdf/ncgen/close.c
```diff
@@ -5,9 +5,7 @@
 
 #include <stdio.h>
 #include "ncgen.h"
-#ifdef EIP
```
byrnHDF, line 8: What is this intended to do "genlib.h"?

PR #77, opened 2023-01-10
Remove most instances of register

schwehr:


byrnHDF (approved):

byrnHDF (changes requested):
genlib.h needs to remove #endif from deleted #if lines.

schwehr:
The genlib.h changes aren't register related.  That's me accidentially getting protoize working mixed in.  I will revert.

schwehr:
> I think there were two see line 21

Definitely.  However, I reverted the entire file as I messed up and put in non-register related changes into this PR.

derobins (approved):

PR #84, opened 2023-01-10
putget.c: nc_API: Be explicit about returning FALSE if caller is null

schwehr:
I originally hit this when I updated hdf4 to 4.2.13 back in 2018

### DIFF
--- a/mfhdf/libsrc/putget.c
+++ b/mfhdf/libsrc/putget.c
@@ -81,6 +81,7 @@ static bool_t nc_API(const char *caller)
     nc_api = strstr(caller, "nc");
     if (nc_api == caller)
         return TRUE;
+    return FALSE;
 }
 
 /*


derobins (approved):

derobins:
Is this a bugfix, then? Maybe we should create a release note.

byrnHDF (approved):
I think there are a few functions that fail to set return values!

schwehr:
A release note would be good. 

I don't think I hit any others with my picky compiler, but there might well be others.

 I can't remember exactly how C specifies the return value where there is no specific return statement for a non-void function.  I thought it would be the last value for C89 / C90, but it's a bug when compiling in C99.

- https://learn.microsoft.com/en-us/cpp/c-language/return-statement-c
- https://stackoverflow.com/questions/4260048/c-function-defined-as-int-but-having-no-return-statement-in-the-body-still-compi says:

> The return value in this case, depending on the exact platform, will likely be whatever random value happened to be left in the return register (e.g. EAX on x86) at the assembly level. Not explicitly returning a value is allowed, but gives an undefined value.
>
> In this case, the 14 is the return value from printf.

Sorry, I'm rusty on finding the canonical source for info on the details of C.

derobins:
Ah, okay, I see. We don't need a release note if we're just addressing bad code that doesn't compile.

PR #88, opened 2023-01-10
Yanks HP-9000 code from the library

derobins:


derobins:
There's a few HP-UX ifdefs lurking, but those frequently relate to Itanium so I'll remove those later

byrnHDF (approved):

PR #106, opened 2023-01-13
Removes almost all mentions of C++ and CXX flags from the library

derobins:
Also updates the C++ extern "C" lines to the current standard


derobins:
GitHub appears to be having difficulty downloading the stuff we need. I'll re-run this job later tonight.

byrnHDF (approved):

PR #108, opened 2023-01-13
Replaced #ifdef <machines> with endianess for conversion functions

bmribler:
Tested on Jelly and Macmini-m1

### DIFF
--- a/hdf/src/hconv.h
+++ b/hdf/src/hconv.h
@@ -58,7 +58,7 @@
 /* CONSTANT DEFINITIONS                                                      */
 /*****************************************************************************/
 /* Generally Big-Endian machines */
-#if !defined(INTEL86) && !defined(__x86_64__)
+#ifdef __BIG_ENDIAN__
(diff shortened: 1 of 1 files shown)


derobins (approved):

hdf/src/hconv.h
```diff
@@ -58,7 +58,7 @@
 /* CONSTANT DEFINITIONS                                                      */
 /*****************************************************************************/
 /* Generally Big-Endian machines */
-#if !defined(INTEL86) && !defined(__x86_64__)
+#ifdef __BIG_ENDIAN__
```
byrnHDF, line 61: Is that the same define as CMake? I thought there was a configure variable; H4_WORDS_BIGENDIAN used by both autotools and CMake.

byrnHDF (changes requested):
I think we need to review why that configure check is there (It is in HDF5 too) and maybe use that define.

derobins:
H4_WORDS_BIGENDIAN is, in fact, already used throughout the code. We should probably use that here, too.

byrnHDF (approved):

PR #112, opened 2023-01-13
Removed DATAINFO macros

bmribler:
They were added to not include functions for hmap project at the
    beginning (2010).  This is not longer necessary.

(tested on jelly)


derobins:
There are also some macros that should be removed in mfhdf/libsrc/mfdatainfo.h :)

byrnHDF (approved):

lrknox (approved):

PR #119, opened 2023-01-14
Add regenerated ncgen flex/bison output files + genparser script

derobins:
The flex/bison-generated files are not routinely generated when the input files change. This brings the genparser script from HDF5 to make running flex/bison easier so we start doing this.

I did not change the input files to eliminate warnings. I will do that in a later PR.


derobins:
I did not change any of the input files so the output changes are due to the flex/bison version changes (which itself could fix bugs) or changes to the input files that were made but never propagated since we didn't routinely regenerate the files.

bin/genparser
```diff
@@ -0,0 +1,162 @@
+#!/bin/sh
+#
+# Copyright by The HDF Group.
+# Copyright by the Board of Trustees of the University of Illinois.
```
lrknox, line 4: Just checking:  Do you want to include this Copyright line?

lrknox (approved):

byrnHDF (approved):

bin/genparser
derobins: I'm just copying the HDF4 lines. Do we need to apply the HDF5 rules to HDF4?

bmribler, line 4: You meant the UofI line?

bmribler: > Just checking: Do you want to include this Copyright line?

We need to keep the UofI copyright line if the file was created before 2006.

PR #124, opened 2023-01-15
Major cleanup of machine-dependent code in hdf/src/hdfi.h

derobins:


hdf/src/hextelt.c
```diff
@@ -81,15 +81,15 @@ EXPORTED ROUTINES
 /* Directory separator definitions relating to a path.
  * Note this does not provide a universal way to recognize
  * different path name conventions and translate between them */
-#if defined WIN386 | defined DOS386
+#ifdef H5_HAVE_WIN32_API
```
byrnHDF, line 84: 5 should be 4

hdf/util/he_file.c
```diff
@@ -15,10 +15,9 @@
 /* --- he-file.c  --- file and annotation manipulation routines */
 #include "he.h"
 
-/* get the prototype for the wait() func. */
-#if defined SUN | defined UNIX386
+#ifdef H5_HAVE_SYS_WAIT_H
```
byrnHDF, line 18: 5 should be 4

byrnHDF (approved):

PR #133, opened 2023-01-16
Fixes the xdr_ops struct in xdrposix.c for non-Solaris systems

derobins:


mfhdf/libsrc/xdrposix.c
```diff
@@ -260,23 +260,16 @@ static struct xdr_ops xdrposix_ops = {
...
+     */
     xdrposix_getint, /* deserialize a 32-bit int */
     xdrposix_putint  /* serialize a 32-bit int */
-#else
-#ifdef AIX5L64
-    xdrposix_destroy,  NULL,           NULL,
-    xdrposix_getint,   xdrposix_putint
-#else  /*AIX5L64 */
-    xdrposix_destroy /* destroy stream */
-#endif /*AIX5L64 */
```
derobins, line 279: Until someone gives me access to an AIX machine, we don't explicitly support AIX anymore.

derobins:
Looks like maybe the system XDR is sad? Will investigate and fix.

bmribler (approved):

mfhdf/libsrc/xdrposix.c
schwehr: It wouldn't be speedy, but maybe AIX on QEMU?

https://kwakousys.wordpress.com/2020/09/06/run-aix-7-2-on-x86-with-qemu/

schwehr:
I had to add `#include <stdint.h>` to xdrposix.c to build and I've got more work to do figure out if thiss solves. the issue, but I think it does.

derobins:
> I had to add `#include <stdint.h>` to xdrposix.c to build and I've got more work to do figure out if thiss solves. the issue, but I think it does.

I added inttypes.h to hdfi.h. That doesn't pull in stdint.h?

mfhdf/libsrc/xdrposix.c
derobins: I've thought about using QEMU, mainly for emulating a BE machine if our existing ones die. I haven't thought about the remaining non-Linux, non-BSD/MacOS UNIX flavors, mainly because they've historically had expensive licenses for their compilers.

derobins: "You can create a free account on IBM but you may have to provide details on an existing serial number and Power Series machine model prior to downloading."

Not sure where I would get something like that

PR #137, opened 2023-01-17
Changed VSgetfields to allow NULL buffer

bmribler:
Description:
    Changed VSgetfields back to allow the buffer 'fields' to be
    NULL so that VSgetfields can return the number of fields in
    the vdata and allow application to allocate sufficient space
    for 'fields'.  The previous fix, in 2001, inadvertently disallowed
    that.


byrnHDF (approved):

derobins (approved):

derobins:
Does this need a release note?

PR #142, opened 2023-01-17
ci: add actions that should pass

hyoklee:
Before you make too many changes, test them with these Actions.
Don't release 4.2.16 until they all pass.


.github/workflows/openbsd.yml
```diff
@@ -0,0 +1,50 @@
...
+          shell: bash          
+          run: |
+            uname -a
+            echo $SHELL
+            pwd
+            ls -lah
+            whoami
+            env | sort
+            sudo pkg_add git
+            git clone https://github.com/hyoklee/hdf4
```
derobins, line 42: This should probably be changed to the canonical repository

byrnHDF:
Can we just combine these into two files - one for aarch and one for bsd.

hyoklee:
> Can we just combine these into two files - one for aarch and one for bsd.

@byrnHDF , good idea but will it give 2 status badges?
I split them to have multiple badges and identify build error easily:

https://github.com/hyoklee/hdf4

derobins:
It looks like the Autotools have trouble with ARM. Can we bring over the tests that pass so we're at least checking for that on each commit? We can set up a separate PR for the platforms and compilers that fail.

derobins:
Is the OpenBSD test ever going to finish?

hyoklee:
> Is the OpenBSD test ever going to finish?

Yes, it finishes in 41 minutes if successful.
It will finish in 6 hours if compilation fails.
Please check log for build error.

derobins:
> > Can we just combine these into two files - one for aarch and one for bsd.
> 
> @byrnHDF , good idea but will it give 2 status badges? I split them to have multiple badges and identify build errors easily:
> 
> https://github.com/hyoklee/hdf4

Ideally, the repo should never be in a broken state, so the badges are nice but not really critical. More commonly, we have tests fail on PRs and then we see the failed tests in the PR, not via badges. I'm okay with splitting alternative architecture tests into separate files since they shouldn't duplicate the matrix from main.yml, but let's try to minimize the splitting to avoid duplication.

PR #144, opened 2023-01-17
Update warnings filter for new extlib builds from fetchcontent

byrnHDF:
Filter out CDash warnings for ext libs.
Eliminate undefineds in the hdf/src and mfhdf/src files

### DIFF
--- a/config/cmake/CTestCustom.cmake
+++ b/config/cmake/CTestCustom.cmake
@@ -5,9 +5,12 @@ set (CTEST_CUSTOM_WARNING_EXCEPTION
     ${CTEST_CUSTOM_WARNING_EXCEPTION}
     ".*note.*expected.*void.*but argument is of type.*volatile.*"
     "src.ZLIB.*:[ \t]*warning"
+    "zlib-src.*:[ \t]*warning"
     "warning LNK4197:.*ZLIB-prefix"
+    "szip-src.*:[ \t]*warning"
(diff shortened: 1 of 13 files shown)


config/cmake/CTestCustom.cmake
```diff
@@ -1,19 +1,28 @@
 # CTestCustom.cmake
-set (CTEST_CUSTOM_MAXIMUM_NUMBER_OF_WARNINGS 1500)
+set (CTEST_CUSTOM_MAXIMUM_NUMBER_OF_WARNINGS 3000)
```
derobins, line 2: Ugh. Windows definitely needs a warning reduction pass. Not sure if I'll have time to do that before the release cutoff date, though. I'll try to take a look to see if there's any low-hanging fruit that gets us a big win with little work.

hdf/src/cszip.c
```diff
@@ -1066,6 +1073,13 @@ HCPsetup_szip_parms(comp_info *c_info, int32 nt, int32 ncomp, int32 ndims, int32
     return (ret_value);
 #else
     /* szip not enabled */
+    (void)c_info;
+    (void)nt;
+    (void)ncomp;
+    (void)ndims;
+    (void)dims;
+    (void)cdims;
```
derobins, line 1081: Honestly, I like this scheme better than the ATTR_UNUSED macros in HDF5

derobins (approved):

PR #145, opened 2023-01-17
Fix examples warning and add macro

byrnHDF:
Merge from PR #22


hdf/src/hdf.h
```diff
@@ -138,6 +138,20 @@ typedef intn (*hdf_termfunc_t)(void); /* termination function typedef */
...
+ * If the checked value matches the specific value, print an error message. */
+#define CHECK_NOT_VAL(ret, val, where)                                                                       \
+    do {                                                                                                     \
+        if (ret == val) {                                                                                    \
+            printf("*** ERROR from %s is %ld at line %4d in %s\n", where, (long)ret, (int)__LINE__,          \
+                   __FILE__);                                                                                \
+        }                                                                                                    \
+    } while (0)
+
+/* .................................................................. */
```
derobins, line 153: I don't think we should put this in the public API. I'm normally against duplicating code, but in this case I'd prefer duplicating it in each set of examples.

derobins: In particular, this is going to be visible in the test's tutil.h file, which seems like it has a high probability of causing problems if we were to refactor things.

derobins (changes requested):
I feel like we should not add the macro to the public API. I don't want to have to support that forever and it'd be confusing to anyone working in the tests (and you KNOW someone is going to use it in the tests at some point...)

byrnHDF:
> I feel like we should not add the macro to the public API. I don't want to have to support that forever and it'd be confusing to anyone working in the tests (and you KNOW someone is going to use it in the tests at some point...)

Maybe we come up with something better that can be used public. The idea was to show error checking in the examples.

derobins:
> > I feel like we should not add the macro to the public API. I don't want to have to support that forever and it'd be confusing to anyone working in the tests (and you KNOW someone is going to use it in the tests at some point...)
> 
> Maybe we come up with something better that can be used public. The idea was to show error checking in the examples.

I feel like that's a laudable goal, but I think putting that macro in the public API is asking for trouble.

hdf/src/hdf.h
bmribler, line 153: Why do you want to put it in hdf.h?  I missed this in the conversation.  A header file can be added in the examples directories for anything used in the examples, like in the dumper.

byrnHDF:
Redoing.

PR #150, opened 2023-01-18
Master build workflow update to match style used in hdf5

byrnHDF:


hyoklee (approved):

hyoklee (approved):

hyoklee:
@byrnHDF , can you also add tests for --disable-shared?

derobins (approved):

derobins:
> @byrnHDF , can you also add tests for --disable-shared?

I went ahead and merged this, but agree that we might want to add --disable-shared. I'd do this outside of the matrix, though to avoid the combinatorial explosion. We should probably add one case each for Linux, Windows, and MacOS since they have different binary issues (.so, .dll, dealing with the universal binary ecosystem).

PR #153, opened 2023-01-19
Fixes the configure.ac sed regex to handle the h4config.h endian block

derobins:


derobins:
This fix also needs to go to HDF5, so I'll set up a PR there, too.

configure.ac
```diff
@@ -91,20 +91,11 @@ AC_PREFIX_DEFAULT([`pwd`/hdf4])
 ## configure.
 AC_CONFIG_COMMANDS([h4config], [
   echo "modifying hdf/src/h4config.h"
-  sed 's/#define /#define H4_/' < hdf/src/h4config.h |\
-    sed 's/#undef /#undef H4_/' > h4config
+  sed 's/#\([[:blank:]]\)*define /#\1define H4_/' < hdf/src/h4config.h |\
+    sed 's/#\([[:blank:]]\)*undef /#\1undef H4_/' > h4config
```
derobins, line 95: This looks complicated, but `([[:blank:]])*` matches and captures any whitespace and drops it where the `\1` is on the right-hand side.

bmribler (approved):

PR #156, opened 2023-01-19
Check exceeding nfields max

bmribler:


bmribler:
Jammed with #154 by accident.

PR #159, opened 2023-01-19
Fixed HDFFR-1593

bmribler:
The number of fields in a vdata from the user's file was corrupted and exceeded the maximum allowed.  Continuing the process with it will likely cause segfault at a later time.  This commit added checks for both dumpvd and dumpvg to fail when such number of fields was read from the file.


byrnHDF (approved):

bmribler:
I need to add RELEASE.txt entry.  You can hold off merging this one.

PR #173, opened 2023-01-19
Cleanup warnings in examples by using if-print instead

byrnHDF:


hdf/examples/AN_create_annotation.c
```diff
@@ -28,40 +26,49 @@ main()
     /*
      * Create the HDF file.
      */
-    file_id = Hopen(FILE_NAME, DFACC_CREATE, 0);
+    if ((file_id = Hopen(FILE_NAME, DFACC_CREATE, 0)) == FAIL)
```
bmribler, line 29: I'm not a fan of macros, but for this I could see the advantage of HDF5 doing HERROR macros.  After Hopen fails, only printing something, it'll spit out a chain of errors because of file_id, an_id,... invalid.  You need to terminate the program.  And if the failure is after file is opened, you'll need to close the file, end access,...

hyoklee:
Can we use perror() instead of printf()?

[1] https://en.wikibooks.org/wiki/C_Programming/stdio.h/perror

hdf/examples/AN_create_annotation.c
byrnHDF: The idea is to keep this simple so the user can copy the function without requiring or deciphering macros.
Yes the examples should terminate sooner in most cases - but that is a task for another day. 
Target was just to eliminate warnings.

bmribler: Also, is there a reason we're not using "fprintf(stderr,"?  Joe's idea might also be good.

byrnHDF: Does it matter - the idea is that the examples should just work (no error checking before) And once something fails i(examples have no error management) nothing matters except what statement failed.
So remember, this is to add simple error report which eliminates some warnings and further work is for a different task.

derobins:
> Can we use perror() instead of printf()?
> 
> [1] https://en.wikibooks.org/wiki/C_Programming/stdio.h/perror

perror() is only useful if errno is set

PR #175, opened 2023-01-20
Adds missing callbacks to funclist_t initialization

derobins:


lrknox (approved):

lrknox:
Why did it not run tests?

derobins:
> Why did it not run tests?

It did on the initial commit, but there was a clang-format commit that does not run tests

PR #176, opened 2023-01-20
Reduce easy warnings in the tests

byrnHDF:


mfhdf/test/tattdatainfo.c
```diff
@@ -480,25 +474,25 @@ add_sdsSDG_annotations()
...
         /* write out scientific data set */
-        ret = DFSDadddata(DFAN_SDG_FILE, 2, dimsizes, (VOIDP)data);
+        DFSDadddata(DFAN_SDG_FILE, 2, dimsizes, (VOIDP)data);
 
         { /* write out annotations for 2 out of every 3 */
             refnum = DFSDlastref();
-            ret    = DFANputlabel(DFAN_SDG_FILE, DFTAG_SDG, refnum, labsds);
-            ret    = DFANputdesc(DFAN_SDG_FILE, DFTAG_SDG, refnum, descsds, (int32)HDstrlen(descsds));
+            DFANputlabel(DFAN_SDG_FILE, DFTAG_SDG, refnum, labsds);
+            DFANputdesc(DFAN_SDG_FILE, DFTAG_SDG, refnum, descsds, (int32)HDstrlen(descsds));
```
derobins, line 488: Shouldn't we be checking the return values of these functions?

byrnHDF: Maybe, but someone would need to understand the calls (@bmribler ?) I just removed the unused variable. I can add a check though.

byrnHDF: Actually the bigger problem is this function, add_sdsSDG_annotations, vs the previous function, add_sdsNDG_annotations. The previous has a var num_errs, which gets returned. This function uses the global var, num_errs, and returns 0.

lrknox (approved):

PR #178, opened 2023-01-20
Revert change to delete files with system("rm -f *.hdf *.tmp"); at end of

lrknox:
hdf/test/testhdf.c for non-Windows machines.

### DIFF
--- a/hdf/test/testhdf.c
+++ b/hdf/test/testhdf.c
@@ -268,7 +268,7 @@ main(int argc, char *argv[])
 
     if (CleanUp) {
         MESSAGE(2, printf("\nCleaning Up...\n\n"););
-#ifndef H5_HAVE_WIN32_API 
+#ifndef H5_HAVE_WIN32_API
         system("rm -f *.hdf *.tmp");
 #endif
     }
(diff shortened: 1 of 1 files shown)


hdf/test/testhdf.c
```diff
@@ -268,8 +268,12 @@ main(int argc, char *argv[])
 
     if (CleanUp) {
         MESSAGE(2, printf("\nCleaning Up...\n\n"););
+#if !(defined DOS386 || defined WIN386)
```
derobins, line 271: These aren't defined anymore, so you want:

`#ifndef H5_HAVE_WIN32_API`

as the test.

hdf/test/testhdf.c
```diff
@@ -268,8 +268,12 @@ main(int argc, char *argv[])
...
+        system("rm -f *.hdf *.tmp");
+#else
         remove("*.hdf");
         remove("*.tmp");
+#endif
```
derobins, line 276: We know the else clause doesn't work, so i'd just delete it. On Windows, files just won't be cleaned up until we really fix file cleanup.

derobins (approved):

PR #179, opened 2023-01-21
Brings HDF5 Autotools configure scheme to HDF4

derobins:
This splits the machine and compiler options into separate files. Only MacOs and Linux have been modified since those are the only machines I have available for testing.

Both Apple and Linux now support:
* icc/icx (icx is the new compiler from OneAPI, icc is considered deprecated)
* gcc
* clang

MacOS currently has a known failure w/ Autotools that does not appear w/ CMake. This is under investigation.

The HDF5 warning scheme has not been brought over at this time. Instead, the "generic" warnings that should work with most currently-used compiler versions are simply added to CFLAGS. When these warnings have been cleaned up, we can bring over the HDF5 scheme and its additional warnings.

### DIFF
--- a/config/apple
+++ b/config/apple
@@ -20,11 +20,20 @@
 #       variable is set based on the incoming value of $CC and is only
 #       used within this file.
 
-# Use clang as the default C compiler.
+# The default compiler is `clang'.
 if test "X-$CC" = "X-"; then
     CC=clang
     CC_BASENAME=clang
 fi
+
+# Figure out GNU C compiler flags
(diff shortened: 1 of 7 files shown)


derobins:
The changes look more complicated than they really are:

* The scheme is very similar to what we do in HDF5, but adapted to HDF4, where we don't have common warning files and the configure options are simpler
* All options have been moved to the compiler files
* The machine files just try all the compilers (like in HDF5) and then set version strings based on what compilers matched

lrknox (approved):

PR #183, opened 2023-01-22
Removes the CONSTR + FUNC scheme for HERROR

derobins:
__func__, et al. are available everywhere now, so this scheme is a giant waste of stack space.


derobins:
We can probably drop that stack size change we added to CMake now that we've gotten rid of the many function name arrays

bmribler (approved):

PR #185, opened 2023-01-22
Eliminate self-assign warning by correcting unused params

byrnHDF:


derobins (approved):

derobins:
@schwehr  suggested doing something like this to avoid creating unused symbols:

```
int
foo(int bar, int /*baz*/)
{
    /* do stuff */
}
```

I think that's a bit cleaner, but we'll have to look into how all compilers treat that (especially MSVC). It might also be a little uglier to ifdef around wrt debug vs release, etc.

PR #187, opened 2023-01-22
Cleans up test/buffer

derobins:
* Noise comment reduction
* Clean up warnings
* Set timezone parameter to NULL in gettimeofday


byrnHDF (approved):

hdf/test/buffer.c
```diff
@@ -428,8 +428,7 @@ main(int argc, char *argv[])
     uint32     lmajor, lminor, lrelease;
     char       lstring[81];
 
-    /* Un-buffer the stdout and stderr */
-    setbuf(stderr, NULL);
```
derobins, line 432: stderr is already unbuffered...

PR #194, opened 2023-01-23
Fixed test_extend_SDSs

bmribler:
Adjusted the offset array that's used to verify getting data info test. The actual offsets in the file became different after an uninitialized buffer to SDwritedata was fixed.

This commit also made the tests fail when the offsets and lengths returned incorrect.  That was missing previously.
(tested on jelly-Linux/64)


byrnHDF (approved):

byrnHDF:
Format_check and commit should be executing on your fork - unless you do not have actions enabled?

PR #195, opened 2023-01-23
Converts alloc.h functions to standard C library functions

derobins:
alloc.h remains since it's used in other directories


derobins:
* Casting from malloc/calloc is a C++ thing, not a C thing
* You can call free() on a NULL pointer, so the checks are pointless
* The HD memory allocation functions need to go away

byrnHDF (approved):

PR #197, opened 2023-01-23
Fixes some unused function warnings

derobins:


derobins:
In some cases, I reordered the includes into the usual order

byrnHDF (approved):

PR #201, opened 2023-01-23
Clean up bit-vector code

derobins:
Strips the bit-vector data structure down to its essential core and cleans it up:

* Removes unused functionality
* Changes internal types to reduce casting
* Replaces HDF-isms with modern C
* Removes noise comments
* Adds additional testing
* Eliminates warnings


derobins:
Working my way down the list of files with compile warnings...

byrnHDF (approved):

PR #204, opened 2023-01-23
Adds -std=c99 to clang, gnu, intel configs

derobins:


derobins:
This is just a draft while I get this to work

config/gnu-flags
```diff
@@ -98,7 +98,7 @@ fi
 
 if test "X-gcc" = "X-$cc_vendor"; then
 
-    CFLAGS="$CFLAGS -Wall -Wcast-qual -Wconversion -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Winvalid-pch -Wmissing-include-dirs -Wshadow -Wundef -Wwrite-strings -pedantic -Wno-c++-compat"
+    CFLAGS="$CFLAGS -std=c99 -Wall -Wcast-qual -Wconversion -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Winvalid-pch -Wmissing-include-dirs -Wshadow -Wundef -Wwrite-strings -pedantic -Wno-c++-compat"
```
byrnHDF, line 101: I think it would be best to use the compiler flags files and "add flags" macros from hdf5 so we can be sure that CMake and autotools are using the same flags.

derobins: Doesn't CMake just do the right thing when you set the standard, though?

byrnHDF: Yes, CMake seems to be fine.

byrnHDF (approved):

PR #205, opened 2023-01-23
Move getopt and cleanup obsolete files

byrnHDF:


mfhdf/hdiff/hdiff_main.c
```diff
@@ -16,6 +16,8 @@
 #include <string.h>
 #include <ctype.h>
 #include <stdlib.h>
+
+#include "../libsrc/getopt.h"
```
derobins, line 20: I feel like this should be protected somehow. I'm also not excited about relative paths to headers.

byrnHDF: I know I added that but that probably doesn't need to be there - I will check.

byrnHDF: fixed.

derobins (approved):

PR #214, opened 2023-01-24
Fix Intel [-Wimplicit-function-declaration] error on Windows

byrnHDF:


hdf/util/he_main.c
```diff
@@ -67,6 +67,10 @@
 #ifdef H4_HAVE_UNISTD_H
 #include <unistd.h>
 #endif
+/* NO_GETPID defined if the OS lacks the getpid() function */
+#ifndef NO_GETPID
+pid_t getpid(void);
+#endif /* !NO_GETPID */
```
derobins, line 73: NO_GETPID isn't defined anywhere, so this just creates dead code. getpid() is POSIX, so you can check for H4_HAVE_UNISTD_H, or just remove the ability to emit the pid and simplify.

derobins: For bonus points, clean up NO_GETPID in mfhdf/libsrc/file.c

derobins: Where is NO_GETPID defined?

byrnHDF:
I think I finally got this cleaned up. If CI pass.

mfhdf/libsrc/file.c
```diff
@@ -501,23 +501,14 @@ ncnobuf(int cdfid)
...
-#define TN_NACCES 0
-#endif /* !NO_ACCESS */
-/* NO_GETPID defined if the OS lacks the getpid() function */
-#ifndef NO_GETPID
+#define TN_NACCES  1
 #define TN_NDIGITS 4
 #if defined H4_HAVE_WIN32_API
     typedef int pid_t;
+#else
+    pid_t getpid(void);
```
derobins, line 509: It looks like you are defining a new prototype for getpid() on non-Windows systems? I don't think that's going to be the right thing to do.

byrnHDF: I think it's trying to match the prototypes.  Windows doesn't define pid_t so that is being set to int. Mingw does define pid_t but getpid returns int. *nix just use pid_t.

derobins (approved):

PR #217, opened 2023-01-26
Makes the FreeBSD Autotools config file work like Linux

derobins:
Defaults are cc (was: gcc) and f77

### DIFF
--- a/config/freebsd
+++ b/config/freebsd
@@ -20,109 +20,120 @@
 #       variable is set based on the incoming value of $CC and is only
 #       used within this file.
 
-if test "X-$CC" = "X-"; then
-  CC=gcc
-  CC_BASENAME=gcc
+# The default compiler is `cc', which should map to clang
+if test -z "$CC"; then
+  CC=cc
+  CC_BASENAME=cc
 fi
 
+# Figure out Clang C compiler flags
+. $srcdir/config/clang-flags
+
+# Figure out GNU C compiler flags
+. $srcdir/config/gnu-flags
+
+# Figure out Intel C compiler flags
+. $srcdir/config/intel-flags
+
+# The default compiler is 'f77', which should map to gfortran
 if test "X-$F77" = "X-"; then
   F77=f77
   F77_BASENAME=f77
 fi
 
-# C and Fortran Compiler and Preprocessor Flags
-# ---------------------------------------------------
-#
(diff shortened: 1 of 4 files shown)


derobins:
We already made these changes in HDF5, so they were easy to port here.

release_notes/RELEASE.txt
```diff
@@ -38,6 +38,18 @@ New features and changes
 ========================
     Configuration:
     -------------
+    - The FreeBSD Autotools config was updated to work like MacOS and Linux
+
+      config/freebsd was updated to try a variety of compilers and the
```
bmribler, line 43: Perhaps: "The file config/freebsd" or something...

config/freebsd
```diff
@@ -20,110 +20,121 @@
...
+        fi
+        ;;
+
+     *)
+        echo "No match to get fc_version_info for $F77"
+        ;;
 esac
 
+if test X != "X$fc_version"; then
+    # Get the compiler version numbers
```
bmribler, line 134: number vs. numbers?

config/freebsd
```diff
@@ -20,110 +20,121 @@
...
+        fc_version_info=`$F77 $FFLAGS -V 2>&1 | grep 'Version' |\
+            sed 's/\(Intel.* Compiler\).*\( Version [a-z0-9\.]*\).*\( Build [0-9]*\)/\1\2\3/'`
+        fc_version="`$F77 $FFLAGS -V 2>&1 |grep '^Intel'`"
+        if test X != "X$fc_version"; then
+            fc_version="`echo $fc_version |sed 's/.*Version \([-a-z0-9\.\-]*\).*/\1/'`"
+        fi
+        ;;
+
+    *f95*)
+        # Figure out which compiler we are using: pgf90 or Absoft f95
```
bmribler, line 102: I don't see pgf90

bmribler (approved):

config/freebsd
derobins: I think they say "numbers" because it's often something like "12.1.0" (major.minor.patch)

config/freebsd
derobins: It's copied over verbatim from HDF5. Since I plan to bring over PGI support in a day or three, I figured I'd leave it in.

byrnHDF (approved):

PR #226, opened 2023-01-28
Include h4config.h early in hdf/src/hdfi.h

derobins:


hdf/src/hextelt.c
```diff
@@ -74,7 +74,7 @@ EXPORTED ROUTINES
 
 ------------------------------------------------------------------------- */
 
-#include "hdf.h"
+#include "hdfi.h"
```
derobins, line 77: On my machine, stat() isn't found when building

byrnHDF (approved):

byrnHDF (approved):

PR #227, opened 2023-01-28
Cleans up atom.c/h

derobins:
* Assumes we're always caching atoms
* Removes HD prefixes
* General tidying


derobins:
This accidentally got committed with the revert of the config code *sigh*

PR #230, opened 2023-01-29
Revert the revert

derobins:
I accidentally merged a bunch of atom changes that really should go into a separate commit when I reverted the warning changes. This reverts the (atom + warning) change, putting the warning changes back in, the reverts the warning commit again, so we're just yanking the warnings changes that break the Autotools.


derobins:
omg I hate git so hard sometimes

PR #240, opened 2023-01-29
Add get pid to config check and add ifdef check

byrnHDF:


hdf/util/he_main.c
```diff
@@ -68,15 +68,18 @@
...
 int         getpid(void);
 #else
 #if defined H4_HAVE_WIN32_API
 typedef int pid_t;
+pid_t       _getpid(void);
 #else
 pid_t getpid(void);
 #endif
 #endif
+#endif
```
derobins, line 82: I don't understand why you are declaring a function we don't implement. This isn't like HDF5 where we map HDgetpid() to some other function via macros. The only thing you should need here is `typedef int pid_t` for any Windows-y platform where pid_t doesn't exist. Once you do that, you can just call getpid(). All POSIX platforms should just need to include unistd.h to get everything they need.

Also, we should ignore Microsoft's underscore versions of things like getpid(). Nobody else cares about being that pedantic and MS plans to support the old, normal names for eternity.

PR #243, opened 2023-01-31
Fixes last machine-specific ifdefs

derobins:


hdf/test/fortest.c
```diff
@@ -84,8 +84,7 @@ main(int argc, char *argv[])
    the size of C pointer; this happens on all 64-bit platforms.
    We need a better fix; see HDFFR-191.
 */
-#if defined _WIN64 || defined __x86_64__ || defined _LP64 || defined __LP64__ || defined __powerpc64__ ||    \
-    defined                                                                              __aarch64__
+#if defined _WIN64 || H4_HAVE_LP64
```
derobins, line 87: 64-bit Windows is P64 (longs are 32-bit)

mfhdf/xdr/xdrstdio.c
```diff
@@ -129,15 +129,13 @@ xdrstdio_putlong(XDR *xdrs, const long *lp)
 {
     int32_t mycopy;
 
-#if defined(_LP64)
-    if ((*lp > UINT32_MAX) || (*lp < INT32_MIN))
-        return (FALSE);
-#endif
+    if ((*lp > INT32_MAX) || (*lp < INT32_MIN))
```
derobins, line 132: No need to ifdef this - it's always a constraint of XDR and we can use C99 types to check the size

byrnHDF (approved):

PR #249, opened 2023-02-02
Add const to xdrposix callbacks

derobins:


schwehr:
One less complaint from llvm.

```
third_party/hdf4/mfhdf/libsrc/xdrposix.c:235:5: warning: incompatible function pointer types initializing 'bool_t (*)(struct __rpc_xdr *, const char *, u_int)' (aka 'int (*)(struct __rpc_xdr *, const char *, unsigned int)') with an expression of type 'bool_t (XDR *, const caddr_t, u_int)' (aka 'int (struct __rpc_xdr *, char *const, unsigned int)') [-Wincompatible-function-pointer-types]
    xdrposix_putbytes, /* serialize counted bytes */
    ^~~~~~~~~~~~~~~~~
1 warning generated.
```

From libtirpc/tirpc/rpc/xdr.h:

```c
typedef struct __rpc_xdr {
	enum xdr_op	x_op;		/* operation; fast additional param */
	const struct xdr_ops {
		/* get a long from underlying stream */
		bool_t	(*x_getlong)(struct __rpc_xdr *, long *);
		/* put a long to " */
		bool_t	(*x_putlong)(struct __rpc_xdr *, const long *);
		/* get some bytes from " */
		bool_t	(*x_getbytes)(struct __rpc_xdr *, char *, u_int);
		/* put some bytes to " */
		bool_t	(*x_putbytes)(struct __rpc_xdr *, const char *, u_int);
		/* returns bytes off from beginning */
```

https://github.com/schwehr/libtirpc/blob/14ec9045fe38140f5adbbaced7bdf06284049b42/tirpc/rpc/xdr.h#L112

schwehr:
With cf2882c, this fixed a bunch of my issues.

I think I have issues that are specific to my bazel build that are related to H4_HAVE_LP64.

From hdf/src/hdfi.h

```c
#if LONG_MAX == INT64_MAX
#define H4_HAVE_LP64
#endif
```

So I need to check what LONG_MAX and INT64_MAX are getting set to.

derobins:
Isn't LONG_MAX in limits.h and INT64_MAX in stdint.h on your system?

On my Ubuntu VM, they are defined to:

`#define LONG_MAX 9223372036854775807L`
`#define INT64_MAX      (__INT64_C(9223372036854775807))`

It looks like __INT64_C() pastes LL on the end, which is different, but shouldn't matter for a numerical test.

PR #252, opened 2023-02-04
Quiets 'ar' warnings from -u option

derobins:
The -u option (default in Automake) only updates changed files in the archive. Leaving it out makes packing a bit slower, but ensures a fresh archive every time. The default in GNU ar is to use -D ('deterministic') which creates a fresh archive each time.

It's better for us to use the GNU default of -D than to use the Automake default of -u. Droppig the -u also quiets a complaint from ar that -u is incompatible with -D.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -333,18 +333,34 @@ fi
 AC_PROG_INSTALL
 AC_PROG_LN_S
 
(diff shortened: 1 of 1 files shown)


derobins:
We already made this change in HDF5, btw.

byrnHDF (approved):

PR #256, opened 2023-02-04
Adds CHECK macros to test_file_limits (tfile1.c)

derobins:
This change adds CHECK_VOID macros to inspect the return values from the HDF4 API calls and complain to the test framework when the calls have failed.


derobins:
The testhdf program is complete trash. Look at all the errors that make noise but don't cause the test to fail! The minor change here reports to the error framework instead of stdout, so now the errors are manifest.

These checks MUST pass for real before we release 4.2.16.

byrnHDF (approved):

PR #258, opened 2023-02-04
Correct -Waddress warnings

derobins:
There were several places where the library attempted to perform a NULL pointer check on an array. These were switched to checking for a \0 or NULL pointer in the first element of the array or removed if useless.

### DIFF
--- a/mfhdf/hdiff/hdiff_list.c
+++ b/mfhdf/hdiff/hdiff_list.c
@@ -1183,7 +1183,7 @@ insert_vs(int32 file_id, int32 ref, /* ref of input VS */
     }
 
     /* ignore reserved HDF groups/vdatas; they are lone ones */
-    if (is_lone == 1 && vdata_class != NULL) {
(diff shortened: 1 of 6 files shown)


mfhdf/hdiff/hdiff_list.c
```diff
@@ -1183,7 +1183,7 @@ insert_vs(int32 file_id, int32 ref, /* ref of input VS */
     }
 
     /* ignore reserved HDF groups/vdatas; they are lone ones */
-    if (is_lone == 1 && vdata_class != NULL) {
+    if (is_lone == 1 && vdata_class[0] == '\0') {
```
byrnHDF, line 1186: so vdata_class has to be empty and reserved in order to detach the vdata_id?

derobins: vdata_class is a char array and it's just a check to see if it's empty

byrnHDF (approved):

PR #259, opened 2023-02-04
Cleanup of pre-C89/99 cruft in mfhdf test files

derobins:
Removes hacks for:
* Missing errno
* Missing stdarg.h
* Missing strerror()
* Missing vfprintf()

Code tidying
* Remove void casts for printf(), etc. return values
* Remove calls to fflush stderr, which is already unbuffered
* Remove MDEBUG blocks
* Remove TEST_PC / TEST_WIN blocks
* Remove unused arguments from main()
* Return EXIT_SUCCESS / EXIT_FAILURE
* Rename redefined assert() to cdf_assert()
* Add const to parray() arguments
* Simplify a few global definitions

### DIFF
--- a/mfhdf/libsrc/error.c
+++ b/mfhdf/libsrc/error.c
@@ -19,53 +19,12 @@
  * mechanisms for netcdf
  */
 
-#ifdef NO_STDARG /* The 4.0 release should be ANSI compliant */
-#undef NO_STDARG
-#endif
-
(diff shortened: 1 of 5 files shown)


schwehr (approved):
Great changes!  I don't see any likely trouble spots.  The cleaned up code is easier to read and follow.

byrnHDF (approved):

PR #260, opened 2023-02-05
Remove HD prefix from most memory calls

derobins:
* Removes HD prefix from: memcpy, memset, malloc, realloc, calloc, free
* Replaces Free() and Alloc() w/ malloc/free
* Removes pointless casts, especially in free()
* Removes most checks for NULL before calling free()
* Deletes mfhdf/libsrc/alloc.h

Also did some tidying, removed noise comments, dead code, etc.

HDmemcmp() was left in place for future overlap checks. It will be removed later.


hdf/examples/VG_get_vgroup_info.c
```diff
@@ -84,10 +84,8 @@ main()
             fprintf(stderr, "   Vgroup name %s and class %s\n", vgroup_name, vgroup_class);
             if (Vdetach(vgroup_id) == FAIL)
                 printf("*** ERROR from Vdetach\n");
-            if (vgroup_name != NULL)
-                HDfree(vgroup_name);
-            if (vgroup_class != NULL)
-                HDfree(vgroup_class);
+            free(vgroup_name);
+            free(vgroup_class);
```
derobins, line 88: Freeing a NULL pointer is explicitly allowed by the standard. There is no need to check for NULL.

mfhdf/libsrc/alloc.h
```diff
@@ -1,38 +0,0 @@
...
-/* We need to define these to standard ones when HDF is not defined */
-#ifndef HDF
-#define HDcalloc(nelem, elsize) calloc(nelem, elsize)
-#define HDmemset(dst, c, n)     memset(dst, c, n)
-#define HDrealloc(p, s)         realloc(p, s)
-#define HDmalloc(s)             malloc(s)
-#define HDfree(s)               free(s)
-#endif
-
-#define ARRAYLEN(arr) (sizeof(arr) / sizeof(arr[0]))
```
derobins, line 36: ARRAYLEN is not used anywhere in the code

derobins:
Yeah, this touches a LOT of files, but the changes are pretty straightforward.  The most important thing to check will be anywhere I yanked a NULL check to make sure I didn't remove anything that isn't a simple check for NULL before calling free().

hdf/src/df24f.c
```diff
@@ -104,7 +104,7 @@ nd2igdim(_fcd filename, intf *pxdim, intf *pydim, intf *pil, intf *fnlen)
     if (!fn)
         return (-1);
     ret = DF24getdims(fn, (int32 *)pxdim, (int32 *)pydim, (intn *)pil);
-    HDfree((VOIDP)fn);
+    free(fn);
```
derobins, line 107: I don't know where anyone got the idea that you had to cast a pointer being freed to void * or char *, but man there were a lot of instances of this.

byrnHDF: Probably some arcane compiler warning!

hdf/src/hblocks.c
```diff
@@ -725,10 +721,8 @@ HLIstaccess(accrec_t *access_rec, int16 acc_mode)
     ret_value = HAregister_atom(AIDGROUP, access_rec);
 
 done:
-    if (ret_value == FAIL) { /* Error condition cleanup */
-        if (access_rec->special_info != NULL)
-            HDfree(access_rec->special_info);
-    }
+    if (ret_value == FAIL)
+        free(access_rec->special_info);
```
byrnHDF, line 725: Just standalone view - should "access_rec->special_info" be set to NULL? Not knowing the possible other use for the pointer, might be a good idea to future proof an invalid usage.

hdf/src/hblocks.c
```diff
@@ -1006,14 +998,11 @@ HLIgetlink(int32 file_id, uint16 ref, int32 number_blocks)
 
 done:
     if (ret_value == NULL) { /* Error condition cleanup */
-        if (new_link->block_list != NULL)
-            HDfree(new_link->block_list);
-        if (new_link != NULL)
-            HDfree(new_link);
+        free(new_link->block_list);
```
byrnHDF, line 1001: is new_link guaranteed to be valid after the "done"?

hdf/src/hblocks.c
byrnHDF: See the function: HLPcloseAID

hdf/src/hblocks.c
```diff
@@ -1642,11 +1627,11 @@ HLPcloseAID(accrec_t *access_rec)
         /* free the linked list of links/block tables */
         for (t_link = info->link; t_link; t_link = next) {
```
byrnHDF, line 1628: info here is set to "access_rec->special_info" which could freed in HLIstaccess failure.

hdf/src/hbuffer.c
```diff
@@ -515,12 +515,12 @@ HBPcloseAID(accrec_t *access_rec)
         } /* end if */
 
         /* Free the memory buffer */
-        HDfree(info->buf);
+        free(info->buf);
```
byrnHDF, line 518: That "access_rec->special_info" pointer is used everywhere!

hdf/src/hchunks.c
```diff
@@ -1270,11 +1249,10 @@ HMCIstaccess(accrec_t *access_rec, /* IN: access record to fill in */
 #if 0 /* dynamic allocation causes a problem on HPUX, removed for now -GV */
```
byrnHDF, line 1249: Do we want to keep "#if 0" blocks? OR should this be a "#ifndef HPUX" block

hdf/src/hcomp.c
```diff
@@ -1057,8 +1054,7 @@ HCIstaccess(accrec_t *access_rec, int16 acc_mode)
 
 done:
     if (ret_value == FAIL) { /* Error condition cleanup */
-        if (info != NULL)
-            HDfree(info);
+        free(info);
```
byrnHDF, line 1057: info is a local variable alias for access_rec->special_info.  freeing info will free access_rec->special_info and then access_rec->special_info will be invalid. access_rec->special_info should be freed and set to NULL. 
Maybe the alias should not be used for the free elsewhere as well. (see lines 1424-1425)

hdf/src/hcompri.c
```diff
@@ -108,7 +108,7 @@ HRPconvert(int32 fid, uint16 tag, uint16 ref, int32 xdim, int32 ydim, int16 sche
         HGOTO_ERROR(DFE_ARGS, FAIL);
 
     /* allocate special info struct for buffered element */
-    if ((info = HDmalloc((uint32)sizeof(crinfo_t))) == NULL)
+    if ((info = malloc((uint32)sizeof(crinfo_t))) == NULL)
```
byrnHDF, line 111: Possible LEAK if getting access_rec fails. info shoulld be freed under line 127 before ERROR macro is executed.

hdf/src/hcompri.c
```diff
@@ -448,7 +448,7 @@ HRPcloseAID(accrec_t *access_rec)
        If no more references to that, free the record */
 
     if (--(info->attached) == 0) {
-        HDfree(info);
+        free(info);
```
byrnHDF, line 451: We should stop using local alias for access_rec->special_info pointer - it's just a bad smell waiting for a problem. 
This function could just be:
`if (--(access_rec->special_info->attached) == 0)
  free(access_rec->special_info)
return SUCCEED`

hdf/src/hdfalloc.c
```diff
@@ -147,7 +147,7 @@ HDstrdup(const char *s)
         return (NULL);
 
     /* Allocate space */
-    ret = (char *)HDmalloc((uint32)HDstrlen(s) + 1);
+    ret = (char *)malloc((uint32)HDstrlen(s) + 1);
```
byrnHDF, line 150: Is this file even necessary?

hdf/src/hextelt.c
```diff
@@ -342,18 +342,15 @@ HXcreate(int32 file_id, uint16 tag, uint16 ref, const char *extern_file_name, in
         if (access_rec != NULL)
             HIrelease_accrec_node(access_rec);
         if (info != NULL) {
-            if (info->extern_file_name != NULL)
-                HDfree(info->extern_file_name);
-            HDfree(info);
+            free(info->extern_file_name);
+            free(info);
```
byrnHDF, line 346: Another instance of info alias!

hdf/src/hfiledd.c
```diff
@@ -328,7 +326,7 @@ HTPinit(filerec_t *file_rec, /* IN: File record to store info in */
         ndds = MIN_NDDS;
 
     /* allocate the dd block in memory and initialize it */
-    file_rec->ddhead = (ddblock_t *)HDmalloc(sizeof(ddblock_t));
+    file_rec->ddhead = (ddblock_t *)malloc(sizeof(ddblock_t));
     if (file_rec->ddhead == (ddblock_t *)NULL)
```
byrnHDF, line 330: Is cast necessary

hdf/src/mcache.c
```diff
@@ -251,13 +251,12 @@ mcache_open(VOID *key,       /* IN: byte string used as handle to share buffers
 
 done:
     if (ret_value == RET_ERROR) { /* error cleanup */
-        if (mp != NULL)
-            HDfree(mp);
+        free(mp);
         /* free up list elements */
         for (entry = 0; entry < HASHSIZE; ++entry) {
             while ((lp = mp->lhqh[entry].cqh_first) != (VOID *)&mp->lhqh[entry]) {
```
byrnHDF, line 257: How many more casts to (VOID *) are there in basic if compares?

hdf/src/mfanf.c
```diff
@@ -226,7 +226,7 @@ nafannlist(intf *an_id, intf *atype, intf *etag, intf *eref, intf alist[])
 
     /* create annlist with true int32s to maintain compatibility
     ** with machines that allocate less than 32 bits per int. */
```
byrnHDF, line 228: Are there any machines left? I suppose maybe some IoT nodes that can't even hold a HDF file in the first place.

bmribler:
What is the reason for removing HD prefix?  All these years, we had to make sure we used it...  And, HDF5 is still using it.

hdf/src/hfiledd.c
bmribler: I haven't looked but yeah, EVERYTHING was casted, many of which I had no idea why...

hdf/src/mfgr.c
```diff
@@ -1282,7 +1277,7 @@ GRIget_image_list(int32 file_id, gr_info_t *gr_ptr)
         }     /* end if */
     }         /* end for */
 
-    HDfree(img_info); /* free image info structures */
+    free(img_info); /* free image info structures */
```
byrnHDF, line 1280: If this was moved to below "done" then the frees for every error would not be needed.

hdf/src/tbbt.c
```diff
@@ -15,15 +15,15 @@
...
 #define TBBT_INTERNALS
 #include "tbbt.h"
-#define Alloc(cnt, typ) (typ *)HDmalloc((cnt) * sizeof(typ))
-#define Free(x)         (HDfree((VOIDP)x))
 
 #define KEYcmp(k1, k2, a)                                                                                    \
-    ((NULL != compar) ? (*compar)(k1, k2, a) : HDmemcmp(k1, k2, 0 < (a) ? (a) : (intn)HDstrlen(k1)))
+    ((NULL != compar) ? (*compar)(k1, k2, a) : memcmp(k1, k2, 0 < (a) ? (a) : (intn)HDstrlen(k1)))
 
 VOID tbbt1dump(TBBT_NODE *node, intn method);
```
byrnHDF, line 28: Is this VOID macro still necessary?

hdf/src/hblocks.c
bmribler: Right, new_link needs to be checked for NULL first in case allocation failed.

hdf/src/tbbt.c
```diff
@@ -649,16 +649,17 @@ tbbtrem(TBBT_NODE **root, TBBT_NODE *node, VOIDP *kp)
...
     tree->count        = 0;
     tree->fast_compare = fast_compare;
     tree->compar       = cmp;
     tree->cmparg       = arg;
-    return (tree);
+
+    return tree;
 }
 
 #ifdef WASTE_STACK
```
byrnHDF, line 665: This ifdef seems useless now.

hdf/src/vgp.c
```diff
@@ -2891,13 +2882,13 @@ VPshutdown(void)
             HGOTO_ERROR(DFE_INTERNAL, FAIL);
 
         vtree = NULL;
-    } /* end if */
+    }
 
     if (Vgbuf != NULL) {
```
byrnHDF, line 2887: Probably could remove this check.

hdf/src/vio.c
```diff
@@ -213,26 +213,26 @@ VSPhshutdown(void)
...
             vs->next             = NULL;
-            HDfree(vs);
-        } /* end while */
-    }     /* end if */
+            free(vs);
+        }
+    }
 
     /* free buffer */
     if (Vhbuf != NULL) {
```
byrnHDF, line 231: Another check could be removed

hdf/src/hchunks.c
bmribler: At first, I was going to say just remove the "#if 0" blocks completely because HMCIstaccess is using
uint8 c_sp_header[256] = "";
but then I saw c_sp_header being dynamically allocated in HMCcreate also many years ago and also by -GeorgeV, the same author of the above "#if 0" blocks!  So, now I'm going to say switch c_sp_header back to dynamically allocated, i.e., removing the "#if 0"s and the workaround: uint8 c_sp_header[256] = "";

hdf/src/vparse.c
```diff
@@ -154,13 +153,11 @@ scanattrs(const char *attrs, int32 *attrc, char ***attrv)
 intn
 VPparse_shutdown(void)
 {
-    intn ret_value = SUCCEED;
-
     if (Vpbuf != NULL) {
```
byrnHDF, line 156: Another if check

hdf/src/vrw.c
```diff
@@ -64,10 +64,10 @@ VSPshutdown(void)
 
     /* free global buffers */
     if (Vtbuf != NULL) {
```
byrnHDF, line 66: if removal

hdf/src/hchunks.c
byrnHDF: I think we should go with new standards and use ifdef machine for special cases for machines we still support.

hdf/src/hcomp.c
bmribler: In this case, HCIstaccess, it is ok because access_rec->special_info is being created/initialized and, when errors occur, whatever there for access_rec->special_info already will be freed via info.
However, the better way is to use info the whole time and only make access_rec->special_info when successs.

hdf/src/hcompri.c
bmribler: In this particular case, it doesn't matter, but in general, a function needs to know which special info it is going to access: chunkinfo_t, compinfo_t, crinfo_t, extinfo_t,... and it's simpler to use local alias.
Having access_rec->special_info everywhere makes the code lengthy and hard to read, in my opinion.

byrnHDF: But having a generic alias may lead to misunderstandings that may lead to weird errors if not properly handled.
You know some github user trys to help and creates a mess because of not fully reading code.
I like the idea of just setting the access_rec->special_info upon success.

mfhdf/dumper/hdp.h
```diff
@@ -17,7 +17,7 @@
 #include "hdf.h"
 #include "hfile.h"
 
-/* Global Variables (ick) */
+/* Global Variables */
```
byrnHDF, line 20: We should be able to eliminate this global vinit_done as it only exists in two files; hdp_list.c and hdp_util.c

byrnHDF: I think the usage in hdp_util.c could be a simple function call.

byrnHDF (approved):
This has identified other issues to be worked on, but this PR is fine.

hdf/src/mfanf.c
derobins: > What is the reason for removing HD prefix? All these years, we had to make sure we used it... And, HDF5 is still using it.

It's to paper over platform differences and as a way to allow injection of alternative functionality. In practice, they have usually been defined to the standard C call. We're removing it from HDF5 as well so our code looks like normal C and we can stop hassling people about using the prefix. Most other software does not do something like this.

hdf/src/tbbt.c
derobins: I plan to remove it and its friend VOIDP.

hdf/src/hblocks.c
derobins: I think there's a lot of instances where we should be resetting pointers to NULL after free.

hdf/src/df24f.c
derobins: I think both HDF4 and HDF5 contain a lot of unnecessary pedantry that clouds reading.

hdf/src/hchunks.c
derobins: Most of the machine-specific ifdefs are gone now. I left this one alone because I planned to re-enable it. Most "it segfaults on one machine" issues are real memory issues and I suspect we're in better shape now.

hdf/src/hdfalloc.c
derobins: No idea!

hdf/src/tbbt.c
derobins: There are still a bunch of useless ifdefs. We'll address them as time goes on.

derobins:
I will create an issue to come back to the topics that have been discussed here.

PR #264, opened 2023-02-05
Fix a segfault in ncdump when CFLAGS=-std=c99

derobins:
Fundamentally, this is due to strdup() returning a pointer to invalid memory. Unclear why this is the case (perhaps the POSIX level is incorrect?), but using the library's HDstrdup() replacement works.

### DIFF
--- a/mfhdf/ncdump/ncdump.c
+++ b/mfhdf/ncdump/ncdump.c
@@ -232,34 +232,31 @@ pr_att_vals(nc_type type, int len, void *vals)
 }
 
 /*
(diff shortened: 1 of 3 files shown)


derobins:
Aha! It's the feature test macros. I'll add those to configure.ac and CMake when we set up C99.

https://stackoverflow.com/questions/8359966/strdup-returning-address-out-of-bounds

lrknox (approved):

bmribler (approved):

PR #265, opened 2023-02-05
Remove HD prefix from memcpy calls

derobins:


derobins:
FWIW, I added the HDF5 library's overlap detection scheme to HDmemcpy() and detected no errors.

lrknox (approved):

PR #267, opened 2023-02-06
Fix an incorrect memset

derobins:
A memeset used the size of a pointer to a struct instead of the size of the struct itself. Fixes a -Wsizeof-pointer-memaccess warning.

### DIFF
--- a/mfhdf/test/tdatainfo.c
+++ b/mfhdf/test/tdatainfo.c
@@ -75,7 +75,8 @@ typedef struct {
 intn
 alloc_info(t_hdf_datainfo_t *info, uintn info_count, int32 n_dims)
 {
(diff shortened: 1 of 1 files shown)


lrknox (approved):

mfhdf/test/tdatainfo.c
```diff
@@ -75,7 +75,8 @@ typedef struct {
 intn
 alloc_info(t_hdf_datainfo_t *info, uintn info_count, int32 n_dims)
 {
-    memset(info, 0, sizeof(info));
+    memset(info, 0, sizeof(*info));
```
bmribler, line 78: Should it be sizeof(t_hdf_datainfo_t)?

PR #268, opened 2023-02-06
Clean up emalloc code

derobins:
* Use size_t for size instead of int
* Remove now pointless negative size checks
* Remove HDF-specific ifdefs
* exit(EXIT_FAILURE) on failure
* Use NULL instead of 0
* Initialize pointers

### DIFF
--- a/mfhdf/nctest/emalloc.c
+++ b/mfhdf/nctest/emalloc.c
@@ -5,55 +5,42 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+
 #include "error.h"
 #include "emalloc.h"
-#ifdef HDF
-#include "hdf.h"
-#endif
 
-/* check return from malloc */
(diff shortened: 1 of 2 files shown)


derobins:
These functions are ridiculous and should go away, but until they do, they should do the right thing.,

mfhdf/nctest/emalloc.c
```diff
@@ -5,55 +5,42 @@
...
-
-    if (p == 0) {
+    p = (void *)realloc(ptr, size);
+    if (p == NULL) {
         error("out of memory");
-        exit(1);
+        exit(EXIT_FAILURE);
     }
+
     return p;
```
derobins, line 45: This is a lot less complicated than it looks. Just inspect the raw file to see how simple they are now.

byrnHDF (approved):

PR #269, opened 2023-02-06
Fix some ncgen variables to use size_t

derobins:
Code in ncgen is using integers and longs to store values from strlen. These should be size_t.


derobins:
This is a very short PR that's basically committing some in-flight work I was doing. I'll continue once the emalloc changes are in.

byrnHDF (approved):

PR #272, opened 2023-02-06
Restore VOID and VOIDP

derobins:
The VOID and VOIDP typedefs are used by external software, even though they are located in an internal header file. Although no longer in use in the HDF4 library, they have been re-added to the hdfi.h header for the time being.

### DIFF
--- a/hdf/src/hdfi.h
+++ b/hdf/src/hdfi.h
@@ -139,6 +139,13 @@ typedef uint32_t uint32;
 typedef int          intn;
 typedef unsigned int uintn;
 
+/* These are no longer used in the library, but other software uses them */
+#ifndef VOID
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):

bmribler (approved):

schwehr:
breadcrumb: The trouble was seen with hdf-eos in gdal.

derobins:
> breadcrumb: The trouble was seen with hdf-eos in gdal.

Also, the current version of HDF-EOS from here:

https://wiki.earthdata.nasa.gov/display/DAS/Toolkit+Downloads

I'll contact the maintainers to see if they can fix their code. It doesn't look like it's on GitHub.

PR #277, opened 2023-02-07
Correct type usage in ncgen/generate.c

derobins:
Addresses warnings in generate.c and related, mostly due to not using size_t appropriately.

### DIFF
--- a/mfhdf/ncgen/genlib.c
+++ b/mfhdf/ncgen/genlib.c
@@ -52,36 +52,36 @@ va_dcl                                      /* variable number of error args, if
 }
 
 void *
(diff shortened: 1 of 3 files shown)


mfhdf/ncgen/genlib.c
```diff
@@ -52,36 +52,36 @@ va_dcl                                      /* variable number of error args, if
 }
 
 void *
-emalloc(int size) /* check return from malloc */
+emalloc(size_t size)
```
derobins, line 55: emalloc is duplicated in nctest. We should eliminate one (or both!) of them.

lrknox (approved):

PR #278, opened 2023-02-07
Remove work-arounds for no stdarg.h

derobins:
Removes some old pre-ANSI work-arounds that use varargs.h instead of stdarg.h.

### DIFF
--- a/mfhdf/ncdump/dumplib.h
+++ b/mfhdf/ncdump/dumplib.h
@@ -8,28 +8,12 @@
 
 extern char *progname; /* for error messages */
 
-#ifndef EXIT_FAILURE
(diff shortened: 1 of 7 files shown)


mfhdf/ncdump/dumplib.h
```diff
@@ -8,28 +8,12 @@
 
 extern char *progname; /* for error messages */
 
-#ifndef EXIT_FAILURE
-#define EXIT_SUCCESS 0
-#define EXIT_FAILURE 1
-#endif
```
derobins, line 14: Might as well get rid of this while we're tidying

byrnHDF (approved):

PR #282, opened 2023-02-07
Fix leaks in hrepack - from dynamic analysis

byrnHDF:


mfhdf/hrepack/hrepack_list.c
```diff
@@ -386,6 +386,9 @@ list_vg(int32 infile_id, int32 outfile_id, int32 sd_id, int32 sd_out, int32 gr_i
                 printf("Error: Could not get name length for group with ref <%d>\n", ref);
                 goto out;
             }
+
+            if (vg_name != NULL)
+                free(vg_name);
```
derobins, line 391: Freeing a NULL pointer is allowed by the standard. Checks for NULL are just noise.

byrnHDF: Right I knew that!!!

derobins (approved):

PR #285, opened 2023-02-08
Close access to the infile access record

byrnHDF:


hdf/util/hdfpack.c
```diff
@@ -444,6 +444,7 @@ copy_blocks(mydd_t *dd, int32 infile, int32 outfile)
         }
     }
     Hendaccess(outaid);
+    Hendaccess(inaid);
```
bmribler, line 447: Did you notice some of the Hendaccess statements have if (ret == FAIL) but some don't?

PR #287, opened 2023-02-10
Remove getopt checks in Autotools and H4_HAVE_GETOPT

derobins:
The Autotools build files do not build getopt.c (either of two duplicated versions). There is also no need to check for getopt in the list of functions - current POSIX systems have getopt (via unistd.h) and Windows does not. Checking for unistd.h should be enough.


derobins:
Closing - we're probably going to paper over this by always using the HDF4 getopt.

PR #294, opened 2023-02-22
Revert unintentional deletion of packaging command

byrnHDF:


derobins:
Looks like you need COPYING instead of COPYING.txt

bmribler (approved):

PR #297, opened 2023-02-23
Fix HDFFR-1607

bmribler:
Description:
    When application called HXsetdir after a read of an external dataset,
    the new setting of the external directory didn't have any effect because
    the library would just use the currently opened file.

    Added a flag to close the file and reopen at the new location if the
    external directory has been changed.  This fix also causes SDreaddata to
    fail in the case where the new external directory is not where the
    external file resides, previously, it just used the old opened file.
    The library now also shows the effect of calling HXsetdir with a NULL
    parameter, which means the external directory will then be the current
    directory.

    Also added/modified utility functions in the test directory.
Platform tested:
    Linux/64 (jelly)


bmribler:
I'm on it...

byrnHDF (approved):

mfhdf/test/texternal.c
```diff
@@ -676,10 +676,176 @@ test_special_combos()
...
+
+    /* When srcdir is not available, make up a directory to create the external
+       file to complicate the subsequent reads */
+    if (!strcmp(dir_name, "./")) {
+
+        /* Facilitate the removal of the temporary directory later */
+        temp_dir = (char *)malloc(strlen(TMP_DIR) + 1);
+        strcpy(temp_dir, TMP_DIR);
+
+#if defined H4_HAVE_WIN32_API && defined __MINGW32__
```
byrnHDF, line 731: Why only mingw with WIN32 API? Why not just WIN32 API?

byrnHDF: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/mkdir?view=msvc-170

byrnHDF: Another search comment: I had to add the include in the if/endif for MinGW lib: #include <direct.h> /* _mkdir */

byrnHDF (changes requested):
I think the mkdir #if should just be:
#if defined H4_HAVE_WIN32_API

byrnHDF (approved):

bmribler:
Do I need to do anything else on this one?

PR #324, opened 2023-03-03
Fix memory leaks in ncdump

bmribler:


byrnHDF (approved):

derobins (changes requested):
This is not how getopt works. You can't malloc a string inside of it and free it later. That breaks POSIX getopt behavior.

PR #326, opened 2023-03-05
Added system header files for _mkdir and rmdir

bmribler:


mfhdf/test/texternal.c
```diff
@@ -728,8 +728,16 @@ test_change_extdir(void)
         temp_dir = (char *)malloc(strlen(TMP_DIR) + 1);
         strcpy(temp_dir, TMP_DIR);
 
+#ifdef H4_HAVE_SYS_STAT_H
+#include <sys/stat.h>
+#endif
+#ifdef H4_HAVE_SYS_TYPES_H
+#include <sys/types.h>
+#endif
```
derobins, line 736: include statements should go at the top of a file, not at line 700

bmribler: I know, I was just trying, I thought it might not help but wanted to try to see.  However, I just realized I don't know that until it's tested with Intel on Windows.

derobins: Your branch should test this. You don't need to create a PR. If you have to, add your branch to the list of tested branches in .github/workflows/main.yml.

mfhdf/test/texternal.c
```diff
@@ -728,8 +728,16 @@ test_change_extdir(void)
...
+
 #if defined H4_HAVE_WIN32_API
-        command_ret = _mkdir(temp_dir);
+        if ((command_ret = _mkdir(temp_dir)) == FAIL)
+            command_ret = mkdir(temp_dir, 0755);
```
derobins, line 740: I'm confused as to how this could have worked. mdkir is an alias for _mkdir on Windows. There's no way they could take different numbers of parameters.

PR #335, opened 2023-03-09
float32 -> float and float64 -> double

schwehr:
Leaves the typedefs in hdfi.h for external code that might use those.

```shell
find . -name \*.c | xargs perl -pi -e "s/float32/float/g"
find . -name \*.c | xargs perl -pi -e "s/float64/double/g"
find . -name \*.h | xargs perl -pi -e "s/float32/float/g"
find . -name \*.h | xargs perl -pi -e "s/float64/double/g"
```

Then had to fix hdfi.h typedefs and  fmtfloat functions


schwehr:
Just clang-format failed, which looks to be automatically fixed.

schwehr:
doing a close/open to trigger the tests.

byrnHDF:
> doing a close/open to trigger the tests.

Tests can be rerun in the checks section - There should be a button/dropdown "Rerun all jobs"

schwehr:
Thanks for the tip!

> Tests can be rerun in the checks section - There should be a button/dropdown "Rerun all jobs"

schwehr:
Closing since the project is not ready for this. I'm pretty sure this change is safe to do for all use cases except where users of the library are hijacking the meaning of float32 and float64 because they are on an architecture where float isn't 32 bit and double isn't 64 bit. In which case, I'm not sure the existing code has ever actually worked as intended.

PR #347, opened 2023-05-07
Remove the XDRSTDIO ifdefs

derobins:
Since HDF4 always uses the POSIX layer of XDR, these ifdefs are useless (and xdrstdio appears to be less performant anyway).

Also removes the copy of xdrstdio that exists in the libsrc directory.


mfhdf/libsrc/putget.c
```diff
@@ -430,44 +425,14 @@ xdr_NCvbyte(XDR *xdrs, unsigned rem, unsigned count, char *values)
...
-        /*
-         * N.B. : "a file positioning function must be called between
-         * a write and a read or vice versa"
-         *        - limitations of stdio, open for update
-         */
-        if (!xdr_setpos(xdrs, origin))
-            return (FALSE);
-#endif /* XDRSTDIO */
-        /* next op is a get */
+
```
bmribler, line 428: I don't know what the deal is about using xdrstdio.c instead of xdrposix.c, but is that the right thing to do in this PR?

derobins: I seriously doubt this is used by anyone, but we can discuss

byrnHDF (approved):

PR #348, opened 2023-05-07
Clean up warnings in hdf/test/macros

derobins:


hdf/test/macros.c
derobins, line 1: Using C99 format specifiers and size_t for the loop variables cleans up all the warnings in this file

hdf/src/hdfi.h
derobins, line 1: I just cleaned up the grammar in this comment block

byrnHDF (approved):

Issue #355, opened 2023-05-09
SDreaddata function in JNI uses mismatched PIN_/UNPIN_ARRAY macros


byrnHDF:
https://github.com/HDFGroup/hdfview/issues/119 and https://github.com/HDFGroup/hdfview/issues/120 issues expose a JNI implementation error in the float and double SDreaddata functions.
Both functions were using PIN_xxx_ARRAY_CRITICAL macros to get the data buffers but only used UNPIN_xxx_ARRAY macros to release the buffers. The "CRITICAL" versions should not have been used (the integer versions did not use CRITICAL)

hyoklee:
https://github.com/HDFGroup/hdfview/issues/120

PR #358, opened 2023-05-12
Fix GH #355 mismatched macros

byrnHDF:


derobins (changes requested):
We should bring over just the change that is necessary for the patch, not everything in master

byrnHDF:
> We should bring over just the change that is necessary for the patch, not everything in master

spellcheck wouldn't pass until I fixed the issues.

bmribler (approved):

derobins (approved):

PR #366, opened 2023-06-13
Fix H4_DFI_H endif.

sebastic:
NCL failed to build with hdf4 4.2.16:
```
/usr/include/hdf/dfi.h:128:2: error: #endif without #if
  128 | #endif /* H4_DFI_H */
      |  ^~~~~
```
When the `endif` for `APOLLO` is removed to not close `#ifndef H4_DFI_H` too soon a different error occurs:
```
 /usr/include/hdf/dfi.h:53:33: error: two or more data types in declaration specifiers
    53 | #define float32                 float
       |                                 ^~~~~
 /usr/include/hdf/hdfi.h:121:16: note: in expansion of macro 'float32'
   121 | typedef float  float32;
       |                ^~~~~~~
 /usr/include/hdf/hdfi.h:121:1: warning: useless type name in empty declaration
   121 | typedef float  float32;
       | ^~~~~~~
```
Re-instating `#ifdef APOLLO` resolves the issue.

### DIFF
--- a/hdf/src/dfi.h
+++ b/hdf/src/dfi.h
@@ -40,60 +40,6 @@
 #undef DF_STRUCTOK /* leave it this way - hdfsh expects it */
 
 /*--------------------------------------------------------------------------*/
-/*                      Machine dependencies                                */
-/*--------------------------------------------------------------------------*/
-
-#ifdef APOLLO
-#include <sys/file.h> /* for unbuffered i/o stuff */
-#define int8                    char
-#define uint8                   unsigned char
-#define int16                   short int
-#define uint16                  unsigned short int
-#define int32                   long int
-#define uint32                  unsigned long int
-#define float32                 float
-#define DFmovmem(from, to, len) memcpy(to, from, len)
-#ifndef DF_STRUCTOK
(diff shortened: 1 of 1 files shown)


hyoklee (approved):
Thank you for your PR!

bmribler:
I think PR# 98 meant to remove the entire #ifdef APOLLO block but missed the matching #endif.

sebastic:
> I think PR# 98 meant to remove the entire #ifdef APOLLO block but missed the matching #endif.

Thanks sounds very plausible, removing the entire block may be a better fix that my single line change.

derobins:
> > I think PR# 98 meant to remove the entire #ifdef APOLLO block but missed the matching #endif.
> 
> Thanks sounds very plausible, removing the entire block may be a better fix that my single line change.

Yes. That was a mistake on my part. Good catch! Thanks!

If you could remove the entire block, down to 'endif /*APOLLO */' around line 93, that would be better and we can get that in right away.

derobins (approved):

derobins:
Looks like codespell updated. I'll fix that ASAP.

bmribler (approved):

PR #369, opened 2023-06-17
Fixed HDFFR-1609

bmribler:
Description:
    Fixing HDFFR-1607 introduced a bug where the flag that indicates whether
    the directory of the external file has been changed from the previous
    location was not reset properly.
    Reset the flag after building a filename path for an external element.


release_notes/RELEASE.txt
```diff
@@ -53,8 +53,18 @@ Support for new platforms and compilers
 =======================================
 
 
-Bugs fixed since HDF 4.2.15
+Bugs fixed since HDF 4.2.16
```
lrknox, line 56: This should probably be since 4.2.16-1

lrknox (approved):

mfhdf/test/texternal.c
```diff
@@ -845,6 +846,102 @@ test_change_extdir(void)
 
 } /* test_change_extdir */
 
+/********************************************************************
+   Name: test_HDFFR_1609() - tests writing after creating external file
+                             and without specifying external dir
+   Description:
+    The main contents include:
+    - HXsetdir to a location other that the directory where the external file
```
hyoklee, line 854: other *that* -> other **than**?

mfhdf/test/texternal.c
```diff
@@ -845,6 +846,102 @@ test_change_extdir(void)
...
+
+    status = make_sourcepath(dir_name, MAX_PATH_LEN);
+    CHECK(status, FAIL, "make_datafilename");
+
+    /* When srcdir is not available, make up a directory to create the external
+       file to complicate the subsequent reads */
+    if (!strcmp(dir_name, "./")) {
+
+        /* Facilitate the removal of the temporary directory later */
+        temp_dir = (char *)malloc(strlen(TMP_DIR) + 1);
```
hyoklee, line 889: What happens if there's a out of memory error?

hyoklee (changes requested):

release_notes/RELEASE.txt
bmribler: But isn't this patch 4.2.16-1?

lrknox: Maybe this is 4.2.16-2, but you're right, the previous release was version 4.2.16.

1.

hyoklee (approved):

mfhdf/test/texternal.c
```diff
@@ -842,9 +844,104 @@ test_change_extdir(void)
...
+    char *temp_dir    = NULL;     /* temp dir to create the external file in */
+    int   command_ret = 0;        /* retvalue from system commands */
+    intn  status      = 0;
+    intn  num_errs    = 0; /* number of errors in compression test so far */
+
+    status = make_sourcepath(dir_name, MAX_PATH_LEN);
+    CHECK(status, FAIL, "make_datafilename");
+
+    /* When srcdir is not available, make up a directory to create the external
+       file to complicate the subsequent reads */
```
byrnHDF, line 885: complicate?

byrnHDF (approved):

mfhdf/test/texternal.c
bmribler: It means to say that instead of simply looking for the file in the current directory, the test tries to look elsewhere.  Any suggestion for better words?

byrnHDF: or "allow"

derobins (approved):

PR #374, opened 2023-06-18
Fix gfortran version check

derobins:
The config/gnu-flags code checked for the gfortran version using
a variable that was never set, so -fallow-argument-mismatch was
never set for gcc 10+, which prevents building the Fortran
libraries.

Fixes HDFFR-1582


byrnHDF (approved):

derobins:
Ooops. Accidentally did this on top of the atom changes.

PR #378, opened 2023-06-18
Build Fortran in GitHub actions

derobins:
Only on Linux w/ static


derobins:
This should also go into the 4.2.16 patch release

derobins:
I'll fix this after I remove that ridiculous -Werror scheme that breaks everything in old code

PR #379, opened 2023-06-18
Move all 'warnings as errors' to warnings

derobins:
This scheme was brought over from HDF5 and only worked in CMake. It breaks Fortran, though, since some of those files are *really* old, so all the 'warnings as errors' are now just regular warnings.

Also brings the warnings scheme into the Autotools (gcc only)


derobins:
This is needed to test Fortran (in a later PR)

lrknox (approved):

PR #380, opened 2023-06-18
Remove duplicated CFLAGS

derobins:


hyoklee (approved):

derobins:
> I don't think these are duplications. The effect of the changes is to throw away the initial CFLAGS configuration, so for GCC 10.2.0 in production mode, "CFLAGS: -std=c99 -Wall -Wcast-qual -Wconversion -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Winvalid-pch -Wmissing-include-dirs -Wshadow -Wundef -Wwrite-strings -pedantic -Wno-c++-compat -Wno-error=implicit-function-declaration -O3 -fomit-frame-pointer" becomes "CFLAGS: -O3 -fomit-frame-pointer" after the change. For the oldest compilers, for loops with declarations will fail to compile, but newer compilers will lose most of the intended compiler flags.

From configure.ac: you can see that CFLAGS is set to CFLAGS and PROD_CFLAGS, etc. Adding CFLAGS to PROD_FLAGS in gnu-flags, etc. duplicates CFLAGS. (I noticed that -std=c99 was showing up twice in the flags list)

```
case "X-$enable_production" in                                                   
  X-|X-yes)                                                                      
    AC_MSG_RESULT([production])                                                  
    CONFIG_MODE=production                                                       
    CFLAGS="$CFLAGS $PROD_CFLAGS"                                                
    FFLAGS="$FFLAGS $PROD_FFLAGS"                                                
    CPPFLAGS="$CPPFLAGS -DNDEBUG $PROD_CPPFLAGS"                                 
    ;;                                                                           
  X-no)                                                                          
    AC_MSG_RESULT([development])                                                 
    CONFIG_MODE=development                                                      
    CFLAGS="$CFLAGS $DEBUG_CFLAGS"                                               
    FFLAGS="$FFLAGS $DEBUG_FFLAGS"                                               
    CPPFLAGS="$CPPFLAGS -UNDEBUG $DEBUG_CPPFLAGS"                                
    ;;                                                                           
  *)                                                                             
    AC_MSG_RESULT([user-defined])                                                
    CONFIG_MODE="user-defined (most CFLAGS, etc. set by user)"                   
    ;;                                                                           
esac 
```

PR #384, opened 2023-06-19
Add Fortran to CI

derobins:


hyoklee (approved):

derobins:
The Autotools has a lot of linking problems. It works fine on my machine, but when Fortran is enabled there are XDR linking problems, regardless of whether we use TI-RPC or our own implementation. For now, I'll just disable it and Fortran will only be tested on CMake until I can figure it out.

PR #388, opened 2023-06-20
Specify compiler in project command

byrnHDF:


lrknox (approved):
CMake configure no longer lists a CXX compiler nor detects iCXX compiler info and features.

PR #391, opened 2023-06-20
Add release workflow, refactor existing scripts

byrnHDF:
Created for preview of workflow for releases.


.github/workflows/release-files.yml
```diff
@@ -0,0 +1,137 @@
...
+          fetch-depth: 0
+
+      - run: |
+          git checkout ${{ inputs.file_sha }}
+
+      - uses: rickstaa/action-create-tag@v1
+        id: "tag_create"
+        with:
+          commit_sha: ${{ inputs.file_sha }}
+          tag: "${{ inputs.use_tag }}"
```
lrknox, line 57: inputs.use_tag means this will ask [a user] to input the release tag?
current formats for release tags:
    hdf-4_vminor_vrelease[-vpatch] for HDF4
    hdf5-1_vminor_vrelease[-vpatch]for HDF5
tag should be applied when release is official and certain, especially so if the tag is public as soon as it is created.  Similarly, release files should only be publicly available when release is final.  Any change after files are put in public release location will require a patch release.

lrknox:
Some additional release tasks that could/should be automated:

- Add contents of RELEASE.txt to HISTORY file.

[ ] HISTORY file has several components:

1. History file header - ~15 lines
2. List of releases.  Latest release is added to the top of the list
3. copies of (entire) RELEASE.txt from each version.  Latest RELEASE.txt is added after the list and before the previous RELEASE.txt, preceded by "%%%4.2.16%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%", followed by 2 blank lines and the previous RELEASE.txt contents. 

- Replace RELEASE.txt in the associated development branch with a saved original with no entries.

- update so numbers.  This may be outside the github release process, but should be done in the associated development branch immediately prior to the creation of the release branch, assuming no further API changes are planned for the release.  Any unplanned API changes would necessitate reviewing the so numbers, and any changes to the release branch numbers should also be made in the development branch.

hyoklee (approved):

hyoklee (approved):

hyoklee (approved):

hyoklee (approved):

.github/workflows/release-files.yml
byrnHDF: Right - the use of inputs prevents the script from running automatically.

lrknox (approved):

PR #392, opened 2023-06-21
Remove dead code behind #ifdefs

derobins:
* #ifdef LATER
* #ifdef DONT_TEST_IMCOMP


derobins:
Tests are failing because GitHub is having trouble finding a place to get MinGW, not because there's a code problem

bmribler (approved):

PR #394, opened 2023-06-21
Clean up INT32DECODE macro

derobins:
The existing macro raises -Wduplicated-branches warnings. This changes adds some casts and other type into to help the compiler understand what's going on, clearing out all the warnings.


derobins:
This is (basically) identical to the corresponding line in the HDF5 macro

byrnHDF (approved):

PR #395, opened 2023-06-21
Remove significant dead code

derobins:
Removes code hiding behind the following unused #ifdefs:
* NOT_YET
* NOT_WORKING
* NOT_IMPLEMENTED
* PERM_OUT
* UNUSED
* NOTUSED
* NO_MULTI_OPEN
* STDIO_BUF
* __CRAY_XT3__


hdf/src/dfstubs.c
```diff
@@ -87,12 +87,6 @@
  *  Important Internal Variables
  */
 static DF *DFlist = NULL; /* pointer to list of open DFs */
-#ifdef PERM_OUT
-static int            DFinuse    = 0;    /* How many are currently in use */
-static uint16         DFmaxref   = 0;    /* which is the largest ref used? */
-static unsigned char *DFreflist  = NULL; /* list of refs in use */
-static char           patterns[] = {0x80, 0x40, 0x20, 0x10, 0x08, 0x04, 0x02, 0x01};
-#endif /* PERM_OUT */
```
derobins, line 95: This looks like the dregs from when DFlist maintained a list of open DFs. Now, despite the comment, it looks like it stores a single int32 atom handle. This generates a LOT of cast alignment warnings that will be addressed later.

Also, FWIW, this part of the library is embarrassing. I can believe people were paid to write this garbage.

byrnHDF (approved):

PR #396, opened 2023-06-22
Fix "failing" netCDF tests

derobins:
A couple of tests in tncvarget would emit messages about reading bad values, but would not actually fail the test. This change fixes all the -Wshadowed warnings, fixes the array code so the tests pass, and properly check the number of errors so the tests will fail on errors.

Note that the library did not actually do the wrong thing - these were just poorly-coded tests.

### DIFF
--- a/mfhdf/test/tncvargetfill.c
+++ b/mfhdf/test/tncvargetfill.c
@@ -141,6 +141,8 @@ read_verify_nc_api_1dim(void)
     }
 
     ncclose(ncid);
+
+    return num_errs;
 } /* end read data */
 
 static int
--- a/mfhdf/test/tncvargetfill.c
+++ b/mfhdf/test/tncvargetfill.c
@@ -60,6 +60,89 @@
 #define SDS2_NAME  "Variable 2"
 #define D1_NUMRECS 8
 #define D2_NUMRECS 2
+
+/*
+ * Helper function: Read and verify data of the variables with nc API
+ */
+static int
+read_verify_nc_api_1dim(void)
+{
+    long  start[1];                                             /* where to start reading */
+    long  edges[1];                                             /* length of data to be read */
+    int   ncid;                                                 /* file id */
+    int   var1id, var2id;                                       /* variable ids */
+    long  dimsize = 0;                                          /* dimension size buffer */
+    short outdata[DIM0];                                        /* data read back */
+    int   rh_ndims;                                             /* number of dims */
+    int   rh_dims[H4_MAX_VAR_DIMS];                             /* variable shape */
+    char  varname[H4_MAX_NC_NAME];                              /* variable name */
+    int16 ncresult1[] = {300, 301, 302, 303, -1, -1, 306, 307}; /* same as sd result */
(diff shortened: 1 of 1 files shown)


mfhdf/test/tncvargetfill.c
```diff
@@ -60,6 +60,89 @@
 #define SDS2_NAME  "Variable 2"
 #define D1_NUMRECS 8
 #define D2_NUMRECS 2
+
+/*
+ * Helper function: Read and verify data of the variables with nc API
+ */
```
derobins, line 66: I separated a couple of functions out to avoid the shadowing warnings when the copy-pasted variables in the blocks shadow the variables in the function body.

mfhdf/test/tncvargetfill.c
```diff
@@ -260,6 +270,143 @@ test_1dim_multivars()
...
+        {{303,  -3}, { -3,  -3}, { -3,  -3}},
+        {{ -3,  -3}, { -3,  -3}, { -3,  -3}},
+        {{ -3,  -3}, { -3,  -3}, { -3,  -3}},
+        {{ -3,  -3}, { -3,  -3}, { -3,  -3}},
+        {{800, 801}, {802, 803}, {804, 805}},
+        {{ -3,  -3}, { -3,  -3}, { -3,  -3}},
+        {{ -3,  -3}, { -3,  -3}, { -3,  -3}},
+        {{ -3,  -3}, { -3,  -3}, { -3,  -3}},
+        {{ -3,  -3}, { -3,  -3}, { -3,  -3}}
+    };
```
derobins, line 311: @bmribler - Make sure I didn't do anything wrong here. I'm new to the HDF4 API but it looks like this was what was intended.

mfhdf/test/tncvargetfill.c
```diff
@@ -463,127 +621,11 @@ test_multidims()
...
-
-        /* Verify data */
-        status = memcmp(outdata1D, ncresult1Ddozen, edges[0] * sizeof(int16));
-        VERIFY(status, 0, "memcmp");
-
-        ncclose(ncid);
-    } /* end read data with nc API */
-
-    return 0;
+    return num_errs;
```
derobins, line 628: All the tests in this file now report the number of errors.

mfhdf/test/tncvargetfill.c
```diff
@@ -622,7 +663,7 @@ test_readings(long max_numrecs)
     /* result data to compare against read data */
 
     /* data resulted from reading at start=[4,0,0] for edges=[6,1,1] */
-    int16 result3D_start400_edge611[DIM0][DIM1][DIM2] = {{-3, -3}, {-3, 800}, {-3, -3}};
+    int16 result3D_start400_edge611[3][2][1] = {{-3, -3}, {-3, 800}, {-3, -3}};
```
derobins, line 666: @bmribler - Ditto here - make sure this is what was intended.

byrnHDF (approved):

bmribler (approved):

PR #428, opened 2023-09-05
Rework private header content

derobins:
The HDF4 library headers frequently used a scheme where private information was
protected by `<FOO>_MASTER` ifdefs. Source files would then define `<FOO>_MASTER`
if they needed to use the internal data. This scheme is baroque, difficult to parse, 
creates unnecessary ifdefs, and results in weirdness like defining global variables in
header files.

This change moves most internal content into the source files in which it is used.
In cases where the content is used among several files (e.g., testing), the content
was moved to a new <foo>i.h internal header file instead. The `<FOO>_MASTER`
symbols have been removed.

Also cleans up the header files, removes some unused symbols, and ensures
that HDFLIBAPI is used consistently.

Only affects the header files in hdf/src. mfhdf will be handled in a separate PR.


hdf/src/dfstubs.h
derobins, line 1: This content was moved to dfstubs.c

derobins:
This PR covers a lot of header files. It might be easier to review if you look at the individual commits. I usually just did one header file at a time so those are easier to digest.

hdf/src/atom.c
```diff
@@ -53,11 +53,63 @@ MODIFICATION HISTORY
...
+
+/* # of bits to use for the Atom index in each atom (change if MAXGROUP>16) */
+#define ATOM_BITS 28
+#define ATOM_MASK 0x0FFFFFFF
+
+/* # of previous atoms cached, change inline caching macros (HAatom_object & HAIswap_cache) if this changes */
+#define ATOM_CACHE_SIZE 4
+
+/* Map an atom to a Group number */
+#define ATOM_TO_GROUP(a) ((group_t)((((atom_t)(a)) >> ((sizeof(atom_t) * 8) - GROUP_BITS)) & GROUP_MASK))
```
schwehr, line 73: Maybe for a future PR: Consider converting local `#define` macros to `static inline` functions. I'm pretty sure `inline` came in with C99.

schwehr (commented):
Only did a quick skim, but it looks like a solid improvement. Perhaps there are a few more local variables that can be `static`? However, I didn't check to see if they are in the header and used across the library.

bmribler (approved):

PR #431, opened 2023-09-12
Remove dfi.h from the source

derobins:
*NOTE* This is a deployed header file!

This header file is declared internal, but is unused in the library and is undocumented in the RM, UG, and DG.


derobins:
Some of the macros in this file are used in other, unused macros. I will track those down and remove them as well.

hdf/src/dfcomp.c
```diff
@@ -52,8 +52,7 @@
  *          Note that rseq does its own compression, because that is part of
  *          the interactive color raster protocol
  *          The space needed for compression and decompression can be allocated
- *          statically or dynamically, depending on the DF_DYNAMIC flag, and
- *          for entire image or part of it (reused) depending on availability
+ *          for the entire image or part of it (reused) depending on availability
```
derobins, line 55: It's always dynamic in the code below and DF_DYNAMIC is not used anywhere

hdf/src/dfi.h
```diff
@@ -1,75 +0,0 @@
...
- * the COPYING file, which can be found at the root of the source code       *
- * distribution tree, or in https://support.hdfgroup.org/ftp/HDF/releases/.  *
- * If you do not have access to either file, you may request a copy from     *
- * help@hdfgroup.org.                                                        *
- * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
-
-/*-----------------------------------------------------------------------------
- * File:    dfi.h
- * Purpose: HDF internal header file
- * Invokes: stdio.h, sys/file.h
```
derobins, line 17: Note "internal"...

hdf/src/dfstubs.c
```diff
@@ -48,46 +48,6 @@
...
-    }
-
-#define CKWRITE(x, y, z, f, ret)                                                                             \
-    {                                                                                                        \
-        if (DF_WRITE((char *)x, (int)y, (int)z, f) < 0) {                                                    \
-            DFerror = DFE_WRITEERROR;                                                                        \
-            return (ret);                                                                                    \
-        }                                                                                                    \
-    }
-
```
derobins, line 90: These are also unused and invoke the dfi.h macros. Since they are located in a C file, there is no risk if we remove them.

hdf/src/hfile.h
```diff
@@ -116,7 +116,6 @@ typedef short hdf_file_t;
 #define HI_SEEK(x, y)     mlseek(x, (int32)y, 0)
 #define HI_SEEKEND(x)     mlseek(x, 0L, 2)
 #define HI_TELL(x)        mlseek(x, 0L, 1)
-#define DF_OPENERR(f)     ((f) == -1)
 #define OPENERR(f)        (f < 0)
 #endif /* FILELIB == MACIO */
```
derobins, line 120: I'll remove code protected by MACIO in a separate commit. I don't think we need to support MacOS 9 in 2023...

release_notes/RELEASE.txt
```diff
@@ -47,7 +47,10 @@ New features and changes
       Also added a section to INSTALL_CMake.txt with very basic explanation of the
       process to use CMakePresets.
 
+    - Private headers are no longer deployed
 
+      * dfi.h - Declared private, unused, undocumented in UG/RM/DG, has been
+                removed from the library entirely.
```
derobins, line 53: I'll expand this list in future commits as we move private headers out of the deploy list

bmribler (approved):

PR #434, opened 2023-09-13
Remove atom hash table power-of-2 ifdefs

derobins:
The atom code enforces the hash table size being a power of two, yet ifdefs for the case where the hash size can be any size remain.

This change removes the ifdefs

### DIFF
--- a/hdf/src/atom.c
+++ b/hdf/src/atom.c
@@ -72,13 +72,8 @@ MODIFICATION HISTORY
 /* Map an atom to a Group number */
 #define ATOM_TO_GROUP(a) ((group_t)((((atom_t)(a)) >> ((sizeof(atom_t) * 8) - GROUP_BITS)) & GROUP_MASK))
 
-#ifdef HASH_SIZE_POWER_2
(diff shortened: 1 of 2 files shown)


hdf/src/atom.c
```diff
@@ -146,10 +143,11 @@ HAinit_group(group_t grp,      /* IN: Group to initialize */
     /* Assertion necessary for faster pointer swapping */
     assert(sizeof(hdf_pint_t) == sizeof(void *));
 
-#ifdef HASH_SIZE_POWER_2
+    /* Ensure hash_size is not zero and a power of two */
+    if (hash_size == 0)
+        HGOTO_ERROR(DFE_ARGS, FAIL);
     if (hash_size & (hash_size - 1))
         HGOTO_ERROR(DFE_ARGS, FAIL);
-#endif /* HASH_SIZE_POWER_2 */
```
derobins, line 152: The original test allowed zero to be a valid size, which is probably not desired

lrknox (approved):

PR #435, opened 2023-09-13
Fix stack size warning in cdeflate.c

derobins:
Converts a giant stack array to a dynamically-allocated array.

Also removes these internal symbols from cdeflate.h (which is public):
* DEFLATE_BUF_SIZE
* DEFLATE_TMP_BUF_SIZE

### DIFF
--- a/hdf/src/cdeflate.h
+++ b/hdf/src/cdeflate.h
@@ -27,10 +27,6 @@
 #include "zlib.h"
 #undef zintf
 
-/* Define the [default] size of the buffer to interact with the file. */
-#define DEFLATE_BUF_SIZE     4096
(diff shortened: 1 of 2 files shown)


hdf/src/cdeflate.h
```diff
@@ -27,10 +27,6 @@
 #include "zlib.h"
 #undef zintf
 
-/* Define the [default] size of the buffer to interact with the file. */
-#define DEFLATE_BUF_SIZE     4096
-#define DEFLATE_TMP_BUF_SIZE 16384
```
derobins, line 32: These should not be public

lrknox (approved):

PR #436, opened 2023-09-13
Fix -Wdouble-promotion issues

derobins:


derobins:
Most of this just involved casting floats to double in printf (and related) statements

lrknox (approved):

PR #439, opened 2023-09-13
atom.c: Initial demonstration of combining variable definition and initialization

schwehr:


derobins:
We've talked about doing something like this in HDF5, but have decided to stick with the 'function scope variables at the top' scheme there and we should probably keep that style here, too.

Historically, this was obviously due to C89 requirements, particularly MSVC. There are other key reasons to keep the existing scheme, though:

* The #1 reason IMHO is that it makes it a LOT easier to look at a function and be sure that you are cleaning up allocated resources properly. In HDF5, incorrect resource cleanup has been a frequent source of bugs, especially in error conditions. Putting everything at the top, initializing pointers, etc. to "bad" values like NULL, and then making sure that everything allocated is cleaned up at the end has been a solid paradigm for the library. Spreading declarations throughout a function makes this harder.
* Also, it would take a lot of work to refactor the entire library to the new style. This would probably take a long time and the library would have an ugly mixed style while that work was in progress.

I know there are good reasons for making this sort of change, but I think in our being super careful about resource allocation and cleanup trumps other style arguments.

I could maybe be convinced otherwise, but I'm probably going to decline this shift in style, at least for now.

derobins:
Yeah, I can probably be convinced otherwise. I just know that we have more problems in HDF5 with proper error cleanup than we do with using uninitialized values, etc.

I think one key thing, regardless of how variables are declared, is refactoring large functions. I'm hoping that once we get the code into a more refactorable state we can make progress there and then to-may-to / to-mah-to decisions are less impactful.

PR #440, opened 2023-09-14
Create private hdf/src header list for CMake

derobins:


hdf/src/CMakeLists.txt
```diff
@@ -81,7 +81,6 @@ set (HDF4_HDF_SRC_CHDRS
     ${HDF4_HDF_SRC_SOURCE_DIR}/dfsd.h
     ${HDF4_HDF_SRC_SOURCE_DIR}/dfufp2i.h
     ${HDF4_HDF_SRC_SOURCE_DIR}/dynarray.h
-    ${HDF4_HDF_SRC_SOURCE_DIR}/glist.h
     ${HDF4_HDF_SRC_SOURCE_DIR}/H4api_adpt.h
     #${HDF4_BINARY_DIR}/h4config.h
```
byrnHDF, line 85: Should this be left separate or added to the private list?

derobins: I don't know why it's separate, but it can't be private

PR #443, opened 2023-09-14
hdf/test/*.c: Remove redundant local function prototypes

schwehr:


schwehr:
And I misspelled redundant in one of the commits for extra bonus :|

schwehr:
As discussed out-of-band, this PR does not match the style the HDF4 team wants, so I'm closing the PR.

From my point of view, these extra prototypes do not add any safety to the code and they are maintenance burden.

PR #445, opened 2023-09-15
hdf/test: Remove local prototypes by moving local only functions to before use

schwehr:
This gets rid of a few redundant prototypes.

### DIFF
--- a/hdf/test/an.c
+++ b/hdf/test/an.c
@@ -32,98 +32,6 @@
 #define COLS        10
 #define REPS        32 /*2*/ /* number of data sets to write to file */
 
-static void gen2Dfloat(int height, int width, float *data);
-
-static void genimage(int height, int width, float *data, uint8 *image);
-
-static void check_lab_desc(uint16 tag, uint16 ref, char *label, char *desc);
-
-void
-test_an(void)
-{
-    char   labsds[MAXLEN_LAB], labris[MAXLEN_LAB], descsds[MAXLEN_DESC], descris[MAXLEN_DESC];
-    uint8  pal[768];
-    uint8 *image, *newimage;
-    uint16 refnum;
-    int32  ret;
-    intn   rank;
-    int    j;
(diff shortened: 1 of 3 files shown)


byrnHDF (approved):

derobins:
This is another one of those changes I'm on the fence about. Is this really better? In HDF5 we got sick of fragile function ordering and mandated function prototypes. In the tests it's probably okay since they are usually only used in the main function, but I'm not a hurry to go back to having to reorder unrelated code because I moved something. Maintaining the prototypes also sucks. I'm just not sure which one sucks *more*. Thoughts?

bmribler:
> having to reorder unrelated code because I moved something
My opinion, this one is worse.

schwehr:
For this type of stuff, I'm strongly in favor of removing them. The extra stuff adds up. Especially in testing, if there is an order dependency issue showing up, the compiler will complain and it frequently indicates a bug. Here for example, there is one prototype that I had to keep and it indicates there there is problem with a missing function in the headers for the tests.

bmribler:
> a missing function in the headers for the tests.
Could you give me more info on this missing function?

bmribler:
> > > a missing function in the headers for the tests.
> 
> > Could you give me more info on this missing function?
> 
> This:
> 
> ```c
> extern void test_mgr_dup_images();
> ```
> 
> https://github.com/HDFGroup/hdf4/blob/bb23516f05e764db2e8eeec59a429f1180bbe309/hdf/test/mgr.c#L1057
> 
> should be in a header as the function is actually here:
> 
> https://github.com/HDFGroup/hdf4/blob/bb23516f05e764db2e8eeec59a429f1180bbe309/hdf/test/tdupimgs.c
You're right, it should have been in tproto.h, my mistake.  I'll fix it.

schwehr:
As discussed out-of-band, this PR does not match the style the HDF4 team wants, so I'm closing the PR.

Issue #446, opened 2023-09-15
Duplicate strdup symbol


schwehr:
I'm hitting an error with some sanitizer builds with the 2nd strdup in hdf4 caused by #442 that did `HDstrdup` → `strdup` in hdfalloc.c

https://github.com/HDFGroup/hdf4/blob/7cbfa6f279f65dede5dbd8bd787d93239c0e2741/hdf/src/hdfalloc.c#L141

```
backward reference detected: strdup in third_party/gdal/_objs/port_lib/json_object.pic.o refers to third_party/hdf4/_objs/libdf/hdfalloc.pic.o
```

derobins:
I'll remove that duplicate symbol. I didn't even know that existed. I'll probably also replace HIstrncpy (in a separate PR) while I'm at it.

schwehr:
Also here: https://github.com/HDFGroup/hdf4/blob/bb23516f05e764db2e8eeec59a429f1180bbe309/hdf/src/hproto.h#L327

PR #447, opened 2023-09-23
Remove strdup implementation

derobins:
Every platform we support has strdup, so this can be removed.


derobins:
The priority is "high" because this creates warnings about duplicated symbols.

schwehr:
Thanks! This will fix #446

schwehr (approved):
I deployed this change and it fixed the issue was running into.

schwehr:
Not for this PR, but can `HIstrncpy` also be replaced with systems' [strncpy](https://cplusplus.com/reference/cstring/strncpy/)

derobins:
> Not for this PR, but can `HIstrncpy` also be replaced with systems' [strncpy](https://cplusplus.com/reference/cstring/strncpy/)

Yes, but since there that function drops \0 in the last byte, there's a little extra work to spread that around the library. I figured I'd do that when I got back from vacation since that might open some cans of worms when it comes to inappropriate string handling.

PR #451, opened 2023-09-28
nbit.c: Cleanup test_nbit7 to make sure inbuf is freed. 

schwehr:
A first trial of fix an issue found by Coverity in HDF4. I'm sticking to test files so that it has the least chance of impacting users as I get accustomed to doing fixes for HDF4.

Tightened up the scope of local variables and added `const` when possible.

Especially important is moving the allocations to as late as possible
to reduce the number of locations where they need to be freed.

```
CID 1546820 (https://github.com/HDFGroup/hdf4/pull/2 of 2): Resource leak (RESOURCE_LEAK)
10. leaked_storage: Variable inbuf going out of scope leaks the storage it points to.
```

### DIFF
--- a/hdf/test/nbit.c
+++ b/hdf/test/nbit.c
@@ -124,19 +124,6 @@
 #define NBIT_MASK12A 0x0000001f
 #define NBIT_MASK12B 0xffffffffUL
 
-static void test_nbit1(int32 fid);
-static void test_nbit2(int32 fid);
-static void test_nbit3(int32 fid);
-static void test_nbit4(int32 fid);
-static void test_nbit5(int32 fid);
-static void test_nbit6(int32 fid);
-static void test_nbit7(int32 fid);
-static void test_nbit8(int32 fid);
-static void test_nbit9(int32 fid);
-static void test_nbit10(int32 fid);
-static void test_nbit11(int32 fid);
-static void test_nbit12(int32 fid);
-
 static void
 test_nbit1(int32 fid)
 {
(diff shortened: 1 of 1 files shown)


hdf/test/nbit.c
```diff
@@ -571,54 +558,51 @@ test_nbit6(int32 fid)
...
 test_nbit7(int32 fid)
 {
-    int32      aid1;
-    uint16     ref1;
-    int        i;
-    int32      ret;
-    intn       errors = 0;
-    model_info m_info;
-    comp_info  c_info;
-    uint8     *outbuf, *inbuf;
```
derobins, line 581: I feel like it would be a lot easier to initialize inbuf and outbuf to NULL, then free them unconditionally at the end.

Also, I'd like for the code to look like HDF5 to minimize friction moving between the libraries. In HDF5, all declarations are at the top of the function, things that need to be freed are initialized to "bad" values, and we free things at the end when they are non-bad (if the cleanup code can't accept bad values, otherwise unconditionally).

I know there are arguments to the distributed declaration style, but "everything at the top" is what we decided on in HDF5 a long time ago. It's also a LOT easier to ensure that everything allocated in the function is disposed of at the end. In this style of code, you have to read every line of code to do the same thing and cleaning up multiple allocated resources can get really messy.

hdf/test/nbit.c
```diff
@@ -124,19 +124,6 @@
...
-static void test_nbit3(int32 fid);
-static void test_nbit4(int32 fid);
-static void test_nbit5(int32 fid);
-static void test_nbit6(int32 fid);
-static void test_nbit7(int32 fid);
-static void test_nbit8(int32 fid);
-static void test_nbit9(int32 fid);
-static void test_nbit10(int32 fid);
-static void test_nbit11(int32 fid);
-static void test_nbit12(int32 fid);
```
derobins, line 138: In HDF5, we use static prototypes. It's probably less important in test code, but again, I'm not excited about getting into style fights. It's easiest for HDFG maintenance if the libraries look alike.

schwehr:
As discussed out-of-band, this PR does not match the style the HDF4 team wants, so I'm closing the PR. Feel free to rewrite this fix to the HDF4/HDF5 style if you'd like.

PR #453, opened 2023-09-28
tbv: Cleanup test_1 to make sure the bit vectors are deleted.

schwehr:
- Use different variable names for each sub test
- Combine definition and initialization
- Localize variables
- Make a `request_size` variable to avoid repeating the size of 80000

```
CID 1546559 (https://github.com/HDFGroup/hdf4/issues/1-4 of 4): Resource leak (RESOURCE_LEAK)
19. leaked_storage: Variable b going out of scope leaks the storage it points to.
```

### DIFF
--- a/hdf/test/tbv.c
+++ b/hdf/test/tbv.c
@@ -29,31 +29,34 @@ static void test_4(void);
 static void
 test_1(void)
 {
-    bv_ptr b;
-    int32  size;
-    intn   ret;
-
     MESSAGE(6, printf("Testing basic bit-vector creation & deletion\n"););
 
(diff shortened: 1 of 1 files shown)


schwehr:
As discussed out-of-band, this PR does not match the style the HDF4 team wants, so I'm closing the PR. Feel free to rewrite this fix to the HDF4/HDF5 style if you'd like.

Also, this test would be better if it was split into two, either as two test functions or two scopes inside a single test function. That would be a better way of showing that the two checks are supposed to be completely independent. They should not use the same variables to make that independence blatantly obvious.

PR #464, opened 2023-12-12
Correct compression libs from GIT and JPEG CMake

byrnHDF:


CMakeFilters.cmake
```diff
@@ -127,8 +127,10 @@ if (HDF4_ENABLE_JPEG_LIB_SUPPORT)
     endif ()
   else ()
     set (HDF4_ENABLE_JPEG_LIB_SUPPORT OFF CACHE BOOL "" FORCE)
-    message (FATAL_ERROR " JPEG support in HDF4 was requested but not found")
+    message (FATAL_ERROR " JPEG support in HDF4 was requested and required but not found")
   endif ()
+else ()
+  message (FATAL_ERROR " JPEG support in HDF4 is required")
```
bmribler, line 133: Do you want to mention --with-jpeg flag?

byrnHDF: CMake doesn't understand that flag, so no reason to confuse.

bmribler: Oh, OK, so how does the user fix the issue?

byrnHDF: What issue? This message instance is because they purposely tried to disable jpeg.

lrknox (approved):

PR #467, opened 2024-01-02
Fix some spelling errors

byrnHDF:


bmribler (approved):

vchoi-hdfgroup (approved):

hdf/src/hchunks.c
```diff
@@ -2452,7 +2452,7 @@ NAME
 DESCRIPTION
      Set the maximum number of chunks to cache.
 
-     The values set here affects the current object's caching behaviour.
```
schwehr, line 2455: Not for this pull request, but it would be good in where ever code style is documented to specify that the code base should using American English and not British English.

byrnHDF: Yes, after changing one I realized that the spell-checker did not flag that - only my editor did.

PR #476, opened 2024-02-01
Fix incorrect option mask returned by SDgetcompinfo

bmribler:
A special bit, SZ_H4_REV_2, was used to indicate that the szip info was stored in a new way.  This bit was set in the options_mask field of the szip info struct.  As a result, the value of options_mask became incorrect.

This PR removed the special bit from the options_mask before returning it to the application.  This change does not affect the data format in the file.

(JIRA issue HDFFR-82)


derobins:
Definitely need to un-uncomment the mfhdf/fortran/tszip.f code around line 280

bmribler:
Need to rework

bmribler:
Sorry for the size of this PR.  HDFFR-82 only reported SDgetcompinfo but it turned out GRgetcompinfo and SDgetchunkinfo were affected by the same bug.

byrnHDF (approved):

bmribler:
Additional work for Java and Fortran

PR #477, opened 2024-02-01
Code cleanup - Remove getcompress functions

bmribler:
The functions SDgetcompress/GRgetcompress have incorrect behavior and were deprecated in favor of SDgetcompinfo/GRgetcompinfo several releases ago.  They should not be used anymore.

Removed the internal function HCPgetcompress and replaced the contents of SDgetcompress/GRgetcompress with a call to SDgetcompinfo/GRgetcompinfo.


byrnHDF:
We should take them out of Java with deprecate tag. And just remove the tests.

byrnHDF (approved):

byrnHDF:
> We should take them out of Java with deprecate tag. And just remove the tests.

HCP routines were not added to Java, so this is not an issue

PR #484, opened 2024-02-04
Cleaned up ifdefs in mfhdf

derobins:
* Removed printf debugging
* Removed `#ifdef HDF` almost everywhere since that's #defined in mfhdf.h


derobins:
This will have merge conflicts once the XDR PR goes in

byrnHDF (approved):

PR #489, opened 2024-02-04
Remove man pages

bmribler:
We will try to generate man pages from the docs in doxygen instead.


derobins (approved):

derobins:
@bmribler Please add a RELEASE.txt node

PR #498, opened 2024-02-06
Fix a warning from unused error handling

derobins:


byrnHDF (approved):

schwehr:
Thanks for fixing this!

derobins:
Thanks for pointing it out!

PR #501, opened 2024-02-07
Install examples into docdir

opoplawski:
FWIW - this is a patch we've carried for a long time in Fedora to install the examples to a location in compliance with our packaging standards.


byrnHDF:
This will need to be revisited as the examples subdirs will be removed in favor of the HDF4Examples folder.

byrnHDF:
Closing this as examples were moved into a single folder - the issue remains to install HDF4Examples into the correct location.

PR #513, opened 2024-02-09
sync and reduce examples files

byrnHDF:
The HDF5Examples files are duplicates of those in the library, this syncs the changes made in those files before removing those files from the source. As in HDF5, we will need to add install commands and scripts to autotools.


HDF4Examples/C/AN/h4ex_AN_get_annotation_info.c
```diff
@@ -8,130 +8,102 @@ main()
...
     ann_type annot_type = AN_DATA_DESC; /* annotation to be obtained*/
     uint16   ann_tag, ann_ref,          /* tag/ref number of an annotation */
         vgroup_tag = DFTAG_VG;          /* tag of the vgroup */
-    int ret_val    = FAIL;
 
     /********************** End of variable declaration **********************/
 
     /*
-     * Open the HDF file.
+     * Create the HDF file.
```
bmribler, line 23: This is an open, not crThis is an open, not create. Am I missing something?eate. Am I missing something?

byrnHDF: No probably was right then the merge from library messed it up.

lrknox (approved):

bmribler (approved):

PR #520, opened 2024-02-13
Moves some things from hdfi.h to hdf.h

derobins:
Some things in hdfi.h, like the C99-like types, need to be in the public API and should be in hdf.h.

### DIFF
--- a/hdf/src/hdf.h
+++ b/hdf/src/hdf.h
@@ -17,6 +17,46 @@
 #include "hdfi.h"
 #include "hlimits.h"
 
+/*-------------------------------------------------------------------------
+ * Pre-C99 platform-independent type scheme
+ *
+ * These types were added long before C99 was widely supported (or even
+ * existed). They were formerly mapped to native C types on a machine-specific
+ * basis, but they are now mapped to their equivalent C99 types.
+ *-------------------------------------------------------------------------*/
+
+/* Floating-point types */
+typedef float  float32;
+typedef double float64;
+
+/* Characters */
+typedef char          char8;
+typedef unsigned char uchar8;
+typedef char         *_fcd;
+#define _fcdtocp(desc) (desc)
+
+/* Fixed-width integer types */
+typedef int8_t   int8;
+typedef uint8_t  uint8;
+typedef int16_t  int16;
+typedef uint16_t uint16;
+typedef int32_t  int32;
+typedef uint32_t uint32;
+
+/* Native integer types */
+typedef int          intn;
+typedef unsigned int uintn;
+
+/* These are no longer used in the library, but other software uses them */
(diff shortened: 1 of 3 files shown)


hdf/src/hdfi.h
```diff
@@ -157,9 +120,6 @@ typedef void *VOIDP;
  * Fortran definitions
  *-------------------------------------------------------------------------*/
 
-/* size of INTEGERs in Fortran compiler */
-typedef int intf;
-
 /* Integer that is the same size as a pointer */
 typedef intptr_t hdf_pint_t;
```
derobins, line 124: This one is only internal

hdf/src/hdf.h
```diff
@@ -149,4 +189,37 @@ typedef intn (*hdf_termfunc_t)(void); /* termination function typedef */
...
+
+#define HDstrcat(s1, s2)     (strcat((s1), (s2)))
+#define HDstrcmp(s, t)       (strcmp((s), (t)))
+#define HDstrcpy(s, d)       (strcpy((s), (d)))
+#define HDstrlen(s)          (strlen((const char *)(s)))
+#define HDstrncmp(s1, s2, n) (strncmp((s1), (s2), (n)))
+#define HDstrncpy(s1, s2, n) (strncpy((s1), (s2), (n)))
+#define HDstrchr(s, c)       (strchr((s), (c)))
+#define HDstrrchr(s, c)      (strrchr((s), (c)))
+#define HDstrtol(s, e, b)    (strtol((s), (e), (b)))
```
derobins, line 223: I'm aware of code that uses these API calls so they need to be public until we remove them

hdf/src/hdf.h
```diff
@@ -17,6 +17,46 @@
...
+
+/* These are no longer used in the library, but other software uses them */
+#ifndef VOID
+/* winnt.h defines VOID to `void` via a macro */
+typedef void VOID;
+#endif
+typedef void *VOIDP;
+
+/* size of INTEGERs in Fortran compiler */
+typedef int intf;
```
derobins, line 58: In the public Fortran API

hdf/src/hdf.h
```diff
@@ -17,6 +17,46 @@
 #include "hdfi.h"
 #include "hlimits.h"
 
+/*-------------------------------------------------------------------------
+ * Pre-C99 platform-independent type scheme
+ *
+ * These types were added long before C99 was widely supported (or even
+ * existed). They were formerly mapped to native C types on a machine-specific
+ * basis, but they are now mapped to their equivalent C99 types.
+ *-------------------------------------------------------------------------*/
```
derobins, line 26: These types definitely need to be public

bmribler (approved):

PR #524, opened 2024-02-13
Remove unused Duff's device ifdefs

derobins:


hdf/src/hconv.h
```diff
@@ -21,35 +21,6 @@
 
 #include "hdfi.h"
 
-/* This looks redundant because previously, VMS platform would undefine DUFF. */
-/* Don't know why. -AKC 2014/1/18 */
-#define DUFF
-#ifdef DUFF
-#define DUFF_sb4b
```
derobins, line 28: This is the only one of these symbols that is used, and it's ALWAYS set

byrnHDF (approved):

bmribler (approved):

PR #525, opened 2024-02-13
Remove DFconvert() macro

derobins:
This is unused


hdf/src/dfsd.c
```diff
@@ -4673,7 +4672,6 @@ DFSDwriteslab(int32 start[], int32 stride[], int32 count[], void *data)
     int32 *filedims;   /* tmp array containing the dimensions */
                        /*   of the dataset in the file */
     int32 r_error;     /* flag if an error occurred, */
-                       /*   used by DFconvert macro */
```
derobins, line 4676: I just removed the "used by DFconvert macro" lines. Everything else is reformatting.

hdf/src/dfconvrt.h
derobins, line 1: FUNC_CONV isn't defined anywhere and this file isn't included by any other file, so it's dead code.

bmribler (approved):

PR #531, opened 2024-02-14
Remove TBBT_INTERNALS scheme

derobins:
The TBBT code used a weird scheme where the main data structures were defined differently in the header and source files in an attempt to hide private details. This has been switched to use a PIMPL scheme with properly hidden data instead.


derobins:
There's no logic changes here. I just properly hid the private data inside a void pointer instead of weirdly redefining the tree and node data structures.

PR #533, opened 2024-02-14
Fix jpeg test const issue

derobins:


hdf/test/rig.c
```diff
@@ -23,7 +23,7 @@
 #define JPEGFILE        "tjpeg.hdf"
 #define NONHDF_JPEGFILE "tnonhdf_jpeg.hdf"
 
-static const uint8 jpeg_8bit_orig[JPEGY][JPEGX] = {
+static uint8 jpeg_8bit_orig[JPEGY][JPEGX] = {
```
derobins, line 26: The underlying libjpeg calls aren't const, so there's no point making these arrays const

bmribler (approved):

byrnHDF (approved):

PR #537, opened 2024-02-14
Fix several stack/frame size warnings

derobins:
Converts static 1D arrays to dynamic


byrnHDF (approved):

schwehr:
I am excited to try this patch out. Hopefully it will fix the msan issue that I see

PR #540, opened 2024-02-16
Move XDR license info to COPYING

derobins:


schwehr:
It's great that top level file has all of the licenses, but I'm curious why the file is named `COPYING`. Is there a spec that hdf4 is based on that says to prefer `COPYING` over `LICENSE`? Lately `LICENSE` has seemed like the standard, so I did that here: https://github.com/google/earthengine-catalog/blob/b23e68da4b2f27f254f0218f79e0d50af88afc11/LICENSE

derobins:
> It's great that top level file has all of the licenses, but I'm curious why the file is named `COPYING`. Is there a spec that hdf4 is based on that says to prefer `COPYING` over `LICENSE`? Lately `LICENSE` has seemed like the standard, so I did that here: https://github.com/google/earthengine-catalog/blob/b23e68da4b2f27f254f0218f79e0d50af88afc11/LICENSE

This is just historically been the name of the file (the codebase dates back to the 80s)

PR #544, opened 2024-02-17
Sync Autotools m4 Java files w/ HDF5

derobins:


derobins:
These are just the same files from HDF5. The Autotools don't support Java very well, but we grabbed the latest version of the (deprecated) Java m4 files and updated them to better find Java on MacOS.

byrnHDF (approved):

PR #548, opened 2024-02-17
Fix encoder detection w/ --with-szlib & no paths

derobins:
The Autotools --with-szlib option did not correctly detect the presence of the encoder due to a mis-named variable. When enabled, a missing variable would cause the dumper tests to fail due to picking the non-szip file for output diffs.

This is fixed now, and --with-szlib works as expected.

Also cleans up some szip-related warnings.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -756,12 +756,14 @@ case "X-$withval" in
   X-yes)
     HAVE_SZIP="yes"
     AC_CHECK_HEADERS([szlib.h], [HAVE_SZLIB_H="yes"], [unset HAVE_SZIP])
-    if test "x$HAVE_SZLIB" = "xyes" -a "x$HAVE_SZLIB_H" = "xyes"; then
+    if test "x$HAVE_SZIP" = "xyes" -a "x$HAVE_SZLIB_H" = "xyes"; then
       AC_CHECK_LIB([sz], [SZ_BufftoBuffCompress],, [unset HAVE_SZIP])
     fi
 
     if test -z "$HAVE_SZIP"; then
       AC_MSG_ERROR([couldn't find szlib library])
+    else
+      USE_COMP_SZIP="yes"
     fi
     ;;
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -756,12 +756,14 @@ case "X-$withval" in
...
     AC_CHECK_HEADERS([szlib.h], [HAVE_SZLIB_H="yes"], [unset HAVE_SZIP])
-    if test "x$HAVE_SZLIB" = "xyes" -a "x$HAVE_SZLIB_H" = "xyes"; then
+    if test "x$HAVE_SZIP" = "xyes" -a "x$HAVE_SZLIB_H" = "xyes"; then
       AC_CHECK_LIB([sz], [SZ_BufftoBuffCompress],, [unset HAVE_SZIP])
     fi
 
     if test -z "$HAVE_SZIP"; then
       AC_MSG_ERROR([couldn't find szlib library])
+    else
+      USE_COMP_SZIP="yes"
```
derobins, line 766: If this isn't set, the dumper tests will pick the non-szip test files and fail on a bad diff

mfhdf/test/tszip.c
```diff
@@ -903,8 +900,11 @@ test_szip_chunk_3d()
     int16         fill_value = 0; /* Fill value */
     comp_coder_t  comp_type;      /* to retrieve compression type into */
     comp_info     cinfo;          /* compression information structure */
-    int           num_errs = 0;   /* number of errors so far */
-    int           i, j, k;
+    int16         all_data[SDS_DIM0][SDS_DIM1][SDS_DIM2];
+    int16         out_data[SDS_DIM0][SDS_DIM1][SDS_DIM2];
```
derobins, line 904: This was set as a global for some reason, which resulted in a TON of shadow warnings

lrknox (approved):

bmribler (approved):

PR #549, opened 2024-02-18
Enable szip + encoder by default

derobins:
As in HDF5, now that libaec is easily available, we attempt to build HDF4 using szip with the encoder enabled by default in both the Autotools and CMake.


derobins:
I'll fix this to be tolerant of missing szip

byrnHDF (approved):
As long as the configure stops if it does not find szip.  Otherwise there could be surprises down the line (we do'nt want to have the same issues we had with HDFView!)

PR #557, opened 2024-02-18
Switch from COPYING to LICENSE

derobins:


hyoklee (approved):

LICENSE
byrnHDF, line 1: Given that the filename "COPYING" is distributed by multiple products and used in so many infrastructure scripts, maybe we should leave the extra file around for one release period?

PR #560, opened 2024-02-19
Don't install ninja on Windows GitHub actions

derobins:


derobins:
Chocolatey is having trouble after server maintenance

https://status.chocolatey.org/

Closing

PR #563, opened 2024-02-19
Fix stack overflows on Windows w/ oneAPI and MinGW + MSYS2

derobins:


derobins:
This probably turns a stack problem into a heap problem, but it at least turns it into something we can work with more easily and allows tests to pass.

byrnHDF (approved):

PR #565, opened 2024-02-19
Add options to allow tools type selection and naming

byrnHDF:


derobins:
Ideally, what we want is basically how the Autotools work:

* If you build static, you get tools that are statically linked to the HDF4 libraries and named normally
* If you build shared, you get tools that are dynamically linked to the HDF4 libraries and named normally
* If you build static + shared, you only get dynamically linked tools, which are named normally
* There should be an option to build tools that are statically linked to the HDF4 libraries (and named normally), even when building shared (some people want to be able to copy the tools around)
* No tool should ever be named `foo-static` or `foo-shared`

derobins (approved):

PR #567, opened 2024-02-20
Fix for compiling the buffer test

derobins:
* Use hdfi.h to pick up time headers
* Also fix warnings

### DIFF
--- a/hdf/test/buffer.c
+++ b/hdf/test/buffer.c
@@ -25,29 +25,11 @@
         - Buffer the compressed element and get times for the buffered element.
         - Make a new linked block element and get benchmark times
         - Buffer the linked block element and get times for the buffered element.
-
-    BUGS/LIMITATIONS
-
-    EXPORTED ROUTINES
-
-    AUTHOR
-        Quincey Koziol
-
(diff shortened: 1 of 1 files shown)


hdf/test/buffer.c
```diff
@@ -25,29 +25,11 @@
...
-
  */
 
 #define TESTMASTER
 
-#include "hdf.h"
-#ifdef H4_HAVE_SYS_TIME_H
-#include <sys/time.h>
-#endif
+#include "hdfi.h"
```
derobins, line 32: This was the main fix. Everything else was minor cleanup and warning fixes.

hdf/test/buffer.c
```diff
@@ -272,7 +252,7 @@ read_test(int32 aid)
         }
 
         /* Clear input buffer */
-        memset(in_buf, 0, elemsize);
+        memset(in_buf, 0, (size_t)elemsize);
```
derobins, line 255: The memset calls emitted sign warnings since elemsize is a signed integer

lrknox (approved):

PR #573, opened 2024-02-21
Rename private headers <foo>_priv.h

derobins:
This will make it easy to distinguish public v. private symbols.

Also:
* Clean up Java JNI header usage, use hdf.h vs. hdfi.h
* Remove unused hdf/src/src.inc and mfhdf/fortran/common.inc files
* Removed mfhdfi.h after moving its content to mfhdf.h

### DIFF
--- a/mfhdf/nctest/atttests.c
+++ b/mfhdf/nctest/atttests.c
@@ -16,7 +16,7 @@
 
 #include "testcdf.h" /* defines in-memory test cdf structure */
 #include "add.h"     /* functions to update in-memory netcdf */
-#include "error_priv.h"
+#include "error.h"
 #include "emalloc.h"
 #include "tests.h"
 #include "val.h"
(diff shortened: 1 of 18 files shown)


bin/h4vers
```diff
@@ -168,12 +168,12 @@ die "mutually exclusive options given\n" if $set && $inc;
...
 #print "File array is ", @files;
 
-# Determine file to use as hfile.h, README.md,
+# Determine file to use as hfile_priv.h, README.md,
 # release_notes/RELEASE.txt, configure.in,
 # and config/cmake/scripts/HDF4config.cmake.
 # The paths to README.md, release_notes/RELEASE.txt, configure.ac,
 # and config/cmake/scripts/HDF4config.cmake
-# files are always from the directory two levels above hfile.h.
+# files are always from the directory two levels above hfile_priv.h.
```
derobins, line 176: In my next PR, I plan to put the version macros in hdf.h

java/src/jni/h4jni.h
```diff
@@ -17,9 +17,10 @@
  *
  */
 
-#include <jni.h>
 #include <string.h>
 
+#include <jni.h>
+
```
derobins, line 23: I reordered the headers so they are in OS/C --> JNI --> Local order and put them outside the extern C block

byrnHDF (approved):

PR #583, opened 2024-02-22
Move version symbols to hfile.h

derobins:
Most of hfile.h moved to hfile_priv.h

Still in hfile.h:

- LIBVER_MAJOR
- LIBVER_MINOR
- LIBVER_RELEASE
- LIBVER_SUBRELEASE
- LIBVER_STRING
- LIBVSTR_LEN
- LIBVER_LEN

Everything else has been moved to hfile_priv.h

NOTE that hfile.h is NOT included via hdf.h and must be included separately, which is the same behavior as HDF 4.2.x.


release_notes/RELEASE.txt
```diff
@@ -180,6 +180,23 @@ New features and changes
...
+      LIBVER_MINOR
+      LIBVER_RELEASE
+      LIBVER_SUBRELEASE
+      LIBVER_STRING
+      LIBVSTR_LEN
+      LIBVER_LEN
+
+      Everything else has been moved to hfile_priv.h
+
+      NOTE that hfile.h is NOT included via hdf.h and must be included
```
byrnHDF, line 197: Should the behavior be changed? Would it be something users would always need to do?

byrnHDF (approved):

release_notes/RELEASE.txt
derobins: I thought about it, but I don't like adding the generic name to the namespace. If it were H4_<whatever>, I would have done that, but not with something that could cause conflicts.

PR #584, opened 2024-02-22
Add netCDF license to COPYING

derobins:


byrnHDF (approved):
Are there too many "dash" lines?

PR #587, opened 2024-02-22
Add deb and rpm binaries to snapshots

byrnHDF:


hyoklee (approved):
It would be nice if you can test installation using `sudo dpkg -i *.deb` and `sudo alien -i *.rpm`.

byrnHDF:
> It would be nice if you can test installation using `sudo dpkg -i *.deb` and `sudo alien -i *.rpm`.

That should be added to the bintest workflow next.

lrknox (approved):

PR #603, opened 2024-02-24
Corrected handling of SDgetcompress()

derobins:
SDgetcompress() was deprecated in HDF 4.2.9 (2013) and hidden behind ifdefs controlled by the --enable-deprecated-symbols (Autotools) and HDF4_ENABLE_DEPRECATED_SYMBOLS (CMake) options. These ifdefs were incompletely propagated throughout the library.

In this PR:

* The mfhdf.h header now hides the SDgetcompress() call behind an H4_NO_DEPRECATED_SYMBOLS ifdef (pulled in via h4config.h via hdf.h)
* The Java wrappers now correctly exclude SDgetcompress() when deprecated symbols are disabled
* The libhdf4.settings file now indicates whether HDF4 was built with or without deprecated symbols

### DIFF
--- a/java/src/jni/hdfsdsImp.c
+++ b/java/src/jni/hdfsdsImp.c
@@ -1419,6 +1419,7 @@ done:
     return JNI_TRUE;
 }
 
+#ifndef H4_NO_DEPRECATED_SYMBOLS
 JNIEXPORT jboolean JNICALL
 Java_hdf_hdflib_HDFLibrary_SDgetcompress(JNIEnv *env, jclass clss, jlong sdsid, jobject cinfo)
 {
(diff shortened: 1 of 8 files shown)


java/src/jni/hdfsdsImp.c
```diff
@@ -1441,6 +1442,7 @@ Java_hdf_hdflib_HDFLibrary_SDgetcompress(JNIEnv *env, jclass clss, jlong sdsid,
 done:
     return JNI_TRUE;
 }
+#endif /* H4_NO_DEPRECATED_SYMBOLS */
```
derobins, line 1445: @byrnHDF - Should I exclude it like this or move the ifdef inside the function and have it raise an exception?

bmribler (approved):

PR #610, opened 2024-02-25
Address comments in HCget_config_info()

derobins:
There was some confusion about the proper behavior of COMP_CODE_NONE and the fallthrough behavior of a switch statement. The correct behavior is to set the encoder and decoder, even for 'no compression' since the 'none' callbacks will be used.

### DIFF
--- a/hdf/src/cnone.c
+++ b/hdf/src/cnone.c
@@ -14,9 +14,6 @@
 /*
    cnone.c - HDF none encoding I/O routines
 
-   These routines are only included for completeness and are not
-   actually expected to be used.
-
    None of these routines are designed to be called by other users except
    for the modeling layer of the compression routines.
  */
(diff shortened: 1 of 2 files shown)


hdf/src/cnone.c
```diff
@@ -14,9 +14,6 @@
 /*
    cnone.c - HDF none encoding I/O routines
 
-   These routines are only included for completeness and are not
-   actually expected to be used.
-
```
derobins, line 19: These are used with uncompressed data!

hdf/src/hcomp.c
```diff
@@ -1387,27 +1387,17 @@ HCget_config_info(comp_coder_t coder_type, /* IN: compression type */
...
         case COMP_CODE_IMCOMP: /* IMCOMP no longer supported */
             *compression_config_info = 0;
             break;
-        /* This block doesn't look intentional, for there is no "break;"
-           before case COMP_CODE_RLE:, which means *compression_config_info
-           was reassigned to something else even though it is "case
-           COMP_CODE_NONE:"  When I added "break;" for "case COMP_CODE_NONE:",             some tests failed.
-           It needs to be checked out.-BMR, Jul 16, 2012*/
-        case COMP_CODE_NONE: /* "none" (i.e. no) encoding */
-            *compression_config_info = 0;
```
derobins, line 1400: The 'none' encoder still sets up callbacks and needs to indicate that the encoder and decoder are available.

bmribler (approved):

PR #614, opened 2024-02-26
Improve mfhdf tests

bmribler:
Moved long/cluttered test functions out of the main test file hdftest.c.


mfhdf/test/tattributes.c
```diff
@@ -79,7 +80,7 @@
 #define ATTR_LEN_ZERO 0
 
 static intn
-test_count(void)
+test_count()
```
derobins, line 83: C should have the void keyword

bmribler:
It was the failure like this one, which I've seen it last year and I asked @byrnHDF about it, but we couldn't figure out what it was:

Test  #66: MFHDF_TEST-hdftest ...........................***Failed  Required regular expression not found. Regex=[HDF-SD test passes
]
 
I googled, some people said something about the CMake version.  They reinstalled CMake
and the problem was gone.  Have you heard anything about that?  But all the other t
ests passed!

derobins:
It's weird because it passes everywhere else, the line `set (passRegex "HDF-SD test passes")` is clearly set in CMakeTests.cmake, and you didn't change anything regarding CMake.

derobins (approved):

derobins:
The problem was a missing return statement on a function that returned an int, so apparently on MacOS it returned non-zero trash, which was interpreted as errors.

bmribler:
Thank you!

PR #617, opened 2024-02-26
Fix a 4-byte swap that uses weird code

derobins:
The original code has a bizarre loop that included additional switch targets that could not be reached. This fix makes the 4-byte swap work like the much more reasonable 8-byte swap code in the same file.

### DIFF
--- a/hdf/src/dfkswap.c
+++ b/hdf/src/dfkswap.c
@@ -154,67 +154,13 @@ DFKsb4b(void *s, void *d, uint32 num_elm, uint32 source_stride, uint32 dest_stri
 
     if (fast_processing) {
         if (!in_place) {
-            uint32 n = (num_elm + 7) / 8;
-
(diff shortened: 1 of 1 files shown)


hdf/src/dfkswap.c
```diff
@@ -154,67 +154,13 @@ DFKsb4b(void *s, void *d, uint32 num_elm, uint32 source_stride, uint32 dest_stri
...
-                            source += 4;
-                    } while (--n > 0);
+            for (i = 0; i < num_elm; i++) {
+                dest[0] = source[3];
+                dest[1] = source[2];
+                dest[2] = source[1];
+                dest[3] = source[0];
+                dest += 4;
+                source += 4;
             }
```
derobins, line 164: Changed to match the code in DFKsb8b, below. There's no reason for the weird do loop with targets inside.

byrnHDF (approved):

PR #624, opened 2024-02-26
Update config with nvhpc compile flags

byrnHDF:


config/cmake/HDFCompilerFlags.cmake
```diff
@@ -37,6 +37,55 @@ if(_CLANG_MSVC_WINDOWS AND "x${CMAKE_C_COMPILER_FRONTEND_VARIANT}" STREQUAL "xGN
...
+  endif ()
+
+  # With at least NVHPC 23.5 - 23.9, compiling with -O2 or higher and -DNDEBUG
+  # appears to have issues that manifest in the tests as incorrect metadata
+  # checksums being read or memory being corrupted. Compiling without -DNDEBUG
+  # does not appear to have these issues, but is not ideal due to compiling in
+  # asserts and other library debug code. Compiling with -O1 also does not appear
+  # to have these issues, so set maximum optimization level to -O1 for now until
+  # it can be determined whether these issues are compiler-specific or issues
+  # in the library.
```
derobins, line 57: You shouldn't need this in HDF4. That bug is probably HDF5-specific.

config/nvidia-flags
```diff
@@ -0,0 +1,57 @@
...
+
+# Common NVIDIA flags for various situations
+if test "X-nvc" = "X-$cc_vendor" -o "X-nvcc" = "X-$cc_vendor"; then
+    # Default to C99 standard.
+    CFLAGS="$CFLAGS $arch -c99 -Minform=warn"
+
+    DEBUG_CFLAGS="-g -Mbounds"
+    DEBUG_CPPFLAGS=
+    #PROD_CFLAGS="-fast"
+    PROD_CFLAGS=""           # -fast implies -O2 and -O2+ currently has test failures.
```
derobins, line 42: Probably don't need to sweat higher optimization in the Autotools due to the HDF5 bug, either

byrnHDF: Just keeping it consistent with the othe files

lrknox (approved):

PR #630, opened 2024-02-27
Fix for big-endian systems

derobins:
The hdf.h header was moved to a place where the BE system checks didn't pick up the system type. This change puts it back.

### DIFF
--- a/hdf/src/hdf_priv.h
+++ b/hdf/src/hdf_priv.h
@@ -25,6 +25,10 @@
 /* The library always uses UNIXBUFIO */
 #define FILELIB UNIXBUFIO
 
+/* Common library headers */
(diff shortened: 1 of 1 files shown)


bmribler (approved):

hdf/src/hdf_priv.h
```diff
@@ -25,6 +25,10 @@
 /* The library always uses UNIXBUFIO */
 #define FILELIB UNIXBUFIO
 
+/* Common library headers */
+#include "hdf.h"
+#include "herr_priv.h"
+
```
derobins, line 31: hdf.h brings in the H4_WORDS_BIGENDIAN constant through h4config.h. This must be done *before* we set DF_MT, below.

PR #632, opened 2024-02-27
Fix build mode in Autotools GH actions

derobins:


derobins:
--enable-build-mode is an HDF5 thing. HDF4 uses --enable-production and --disable-production.

PR #634, opened 2024-02-27
Update platforms Tested

byrnHDF:


lrknox:
I'll push more platform updates to this PR today.

hyoklee (approved):

bin/h4vers
```diff
@@ -353,7 +353,7 @@ if ($HDF4CONFIGCMAKE) {
   my $sub_rel_ver_str = (
      $newver[3] eq ""
      ? sprintf("\"%s\"", "")
-     : sprintf("\"%s\"", "-".$newver[3].", currently under development")
```
byrnHDF, line 356: Looks simple enough.

lrknox (approved):

release_notes/RELEASE.txt
```diff
@@ -25,7 +25,6 @@ CONTENTS
 - New features and changes
   -- Configuration
   -- C Library
-- Support for new platforms and compilers
 - Bugs fixed since HDF 4.2.16
```
bmribler, line 28: should be 4.2.16-2 as the one below

lrknox: Hmmm.  There's nothing in the HISTORY.txt file for a 4.2.16-2 release.  I don't remember why there was a 4.2.16-2 patch release, but I wonder if the previous versions in this RELEASE.txt should all be 4.2.16?
   
There are 2 instances of 4.2.16 and one of 4.2.16-2, so I suppose all 3 in this file should be the same.

derobins: The 4.3.0 release should be off the 4.2.16, IIRC. A patch realease technically isn't a real release.

derobins (approved):

PR #637, opened 2024-02-27
Fix version information output for hdfls.

hyoklee:
hdfls prints the old version information:
```
hdfls,  version: 2.0   date: March 1, 1994
```

This PR will match other tools for version information.


byrnHDF (approved):
Wow so this has been wrong for 30 years!

lrknox (approved):

PR #654, opened 2024-03-01
Add Cygwin GitHub Action

hyoklee:
It utilizes my.cdash.org for test result reporting.

### DIFF
--- a/.github/workflows/cwin.yml
+++ b/.github/workflows/cwin.yml
@@ -1,0 +1,32 @@
+name: cwin
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  test:
+    runs-on: windows-latest
+    steps:
+      - name: Set git to use LF
+        run: |
+          git config --global core.autocrlf input
+
+      - uses: actions/checkout@v4
+
+      - name: Install Cygwin
+        uses: cygwin/cygwin-install-action@master
+        with:
+          packages: cmake gcc-fortran libjpeg-devel make ninja zlib-devel
+      - name: Test HDF4
+        run: |
+          export PATH=/usr/bin:$PATH
+          mkdir build
+          cd build
+          which cmake
+          cmake -G Ninja \
+           -D BUILD_SHARED_LIBS:BOOL=OFF \
+           -D CTEST_DROP_SITE_INIT:STRING="my.cdash.org" \
+           -D SITE:STRING=cwin ..
+          ctest -D Experimental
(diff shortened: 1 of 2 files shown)


byrnHDF:
I like the idea, but we need to standardize how we do things.
How long does this run? We should incorporate into the callable workflows and consistent runs.
Why report all pushes to CDash? CI will stop and the developer can see the failures, so no need for CDash.
I could see a daily_build reporting to CDash - too many reports of the same can just clutter the CDash page and blur the impact and/or push important reports to another page.
Need to discuss a plan.

byrnHDF:
Thinking more about the reporting to CDash. I think we should create a "Nightly" workflow that runs only once a day when there have been changes the previous day. Then we can have a standard set of configurations that would run across all the platforms and report to CDash. This would identify any issues with configuration failure and also it would be easy to highlight a platform failure.

byrnHDF:
A nightly build would be easy to add an argument to the existing workflows to indicate that builds should send reports to CDash.

.github/workflows/cwin.yml
```diff
@@ -0,0 +1,32 @@
+name: cwin
+
+on:
+  push:
+    branches: [ master ]
```
derobins, line 5: The on: field needs to look like the other GitHub actions:

on:
  workflow_dispatch:
  push:
  pull_request:
    branches: [ master ]
    paths-ignore:
    - '.github/CODEOWNERS'
    - 'doc/**'
    - 'release_docs/**'
    - 'COPYING'
    - '**.md'

.github/workflows/cwin.yml
```diff
@@ -0,0 +1,32 @@
+name: cwin
```
derobins, line 1: This should have a better name

.github/workflows/cwin.yml
```diff
@@ -0,0 +1,32 @@
...
+          packages: cmake gcc-fortran libjpeg-devel make ninja zlib-devel
+      - name: Test HDF4
+        run: |
+          export PATH=/usr/bin:$PATH
+          mkdir build
+          cd build
+          which cmake
+          cmake -G Ninja \
+           -D BUILD_SHARED_LIBS:BOOL=OFF \
+           -D CTEST_DROP_SITE_INIT:STRING="my.cdash.org" \
```
derobins, line 29: If we're going to integrate CDash and GitHub CI, which I'm not sure we should do, we need to do that systematically and not ad-hoc.

derobins (approved):

PR #663, opened 2024-03-02
Rework atexit handling code

derobins:
Allows removal of over-engineered glist code that can't even be used since function pointers aren't pedantically void pointers.


derobins:
I'll probably move the code to separate hfile_atexit.c/h files in a future PR, but this at least gets rid of the -pedantic complaints about storing function pointers in void pointers.

byrnHDF (approved):

PR #666, opened 2024-03-03
Separate atexit code to new files

derobins:
Also adds tests


derobins:
It was kind of gross having all that atexit stuff in hfile.c, so I split it out into it's own file and added some tests.

byrnHDF (approved):

PR #679, opened 2024-03-04
Do not enable szip for sanitizer runs

byrnHDF:


config/cmake/mccacheinit.cmake
```diff
@@ -16,11 +16,9 @@ set (HDF4_BUILD_UTILS ON CACHE BOOL "Build HDF4 Library Utilities" FORCE)
 
 set (HDF4_BUILD_GENERATORS OFF CACHE BOOL "Build Test Generators" FORCE)
 
-set (HDF4_ENABLE_JPEG_LIB_SUPPORT ON CACHE BOOL "Use JPEG Filter" FORCE)
```
bmribler, line 19: Is this JPEG line removed intentionally?

byrnHDF: yes, it is on by default

derobins (approved):

PR #680, opened 2024-03-04
Correct test commands when used with memchecking

byrnHDF:


mfhdf/hdfimport/CMakeTests.cmake
```diff
@@ -154,9 +154,7 @@ macro (ADD_H4_TEST_OUT resultfile resultcode)
 endmacro ()
 
 macro (ADD_H4_TEST_ED testfile resultfile resultcode)
-  if (HDF4_ENABLE_USING_MEMCHECKER)
-    add_test (NAME HIMPORT-EDIT COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:hdfed> -batch)
-  else ()
+  if (NOT HDF4_ENABLE_USING_MEMCHECKER)
```
derobins, line 157: Why is there an option for this? We use USING_MEMCHECKER in HDF5 because we have our own implementation of a memory allocator via the H5FL free list package. There is nothing like that in HDF4.

byrnHDF: It is connected to a CMake process and variables that indicate the use of analysis tools. It is in hdf5 too, and is not the same as USING_MEMCHECKER.

derobins (approved):

PR #684, opened 2024-03-05
Remove Void type

derobins:
* Remove Void (was: typedef to char *, now just void *)
* Remove NULL casts

### DIFF
--- a/mfhdf/src/nc_priv.h
+++ b/mfhdf/src/nc_priv.h
@@ -45,9 +45,6 @@
 #define NC_DOUBLE_SIZE      8
 #define NC_UNSPECIFIED_SIZE 0
 
-/* ptr argument type in internal functions */
-#define Void char
-
 /*
  * Include HDF stuff
  */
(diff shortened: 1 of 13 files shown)


mfhdf/src/nc_priv.h
```diff
@@ -45,9 +45,6 @@
 #define NC_DOUBLE_SIZE      8
 #define NC_UNSPECIFIED_SIZE 0
 
-/* ptr argument type in internal functions */
-#define Void char
```
byrnHDF, line 49: WOW

byrnHDF (approved):

mfhdf/src/nc_priv.h
derobins: HDF4 is always a journey down memory lane

derobins: At least this one is internal. The VOID macro is in the public API and HDF-EOS2 uses it.

bmribler (approved):

derobins:
Closed in lieu of #686. Turns out Void is being used as a byte pointer, not a "generic object" pointer.

PR #688, opened 2024-03-08
Remove intn and uintn from source

derobins:
These represent native types and have been typedef'd to int and unsigned int for a long time. They still exist in hdf.h so old code will compile.


derobins:
I'll do the other types in future PRs

byrnHDF (approved):

schwehr:
Thank you! This makes the code a lot easier to read.

PR #700, opened 2024-03-10
Re-enable test_sd() in hdftest

derobins:
This test was commented out due to failures when running as root. I've added a new hack that skips the second SDstart() call, which can succeed (considered a failure) when a user has root privileges.

### DIFF
--- a/mfhdf/test/hdftest.c
+++ b/mfhdf/test/hdftest.c
@@ -195,10 +195,7 @@ main(void)
        library is not present or only decoder is available. */
     num_errs = num_errs + test_szip_compression();
 
-    /* NOTE: This test had a history of failing on some systems when the user
-     * is logged in as root. It has been re-enabled and we'll investigate
-     * if it starts failing again.
-     */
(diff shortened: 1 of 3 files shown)


mfhdf/test/hdftest.c
```diff
@@ -199,11 +178,9 @@ main(void)
...
     num_errs = num_errs + test_files();
 
-    /* BMR: Added a test routine dedicated for testing the behavior of
-     * several functions when the SDS has rank=0. (in trank0.c) - 02/4/05 */
-    /* BMR: SDcreate fails on Copper when rank=0.  EP decided to remove
-     * this test until further study can be made on this feature.
-    num_errs = num_errs + test_rank0(); */
+    /* Test the behavior of several functions when the SDS has rank=0 (in trank0.c) */
+    /* FIXME: Currently failing - fix ASAP */
+    /* num_errs = num_errs + test_rank0(); */
```
derobins, line 183: I created an issue for this

bmribler (approved):

PR #706, opened 2024-03-29
Fix off-by-1 max file name len in dfp/dfr8

derobins:
The "last file name" arrays in both dfp.c and dfr8.c can only hold DF_MAXFNLEN - 1 characters, so a file name of of length DF_MAXFNLEN (defined in hlimits.h) would be truncated.

This has little practical effect, as it's just used in a shortcut used when reopening files, but it's incorrect behavior and raises warnings. Other "last file" shortcut code uses correctly-sized arrays.

### DIFF
--- a/hdf/src/dfp.c
+++ b/hdf/src/dfp.c
@@ -35,7 +35,7 @@ static uint16 Writeref = 0;
 static uint16 Refset   = 0; /* Ref of palette to get next */
 static uint16 Lastref  = 0; /* Last ref read/written */
 
-static char Lastfile[DF_MAXFNLEN] = ""; /* last file opened */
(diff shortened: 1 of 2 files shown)


hdf/src/dfp.c
```diff
@@ -276,7 +276,7 @@ DFPnpals(const char *filename)
     }
 
     /* Get space to store the palette offsets */
-    if ((pal_off = (int32 *)malloc(npals * sizeof(int32))) == NULL)
+    if ((pal_off = (int32 *)malloc((size_t)npals * sizeof(int32))) == NULL)
         HGOTO_ERROR(DFE_NOSPACE, FAIL);
```
derobins, line 280: This was the only other warning in this file

bmribler (approved):

PR #710, opened 2024-03-30
Warning cleanup in hdf/src and test & mfhdf/src

derobins:
Mainly cleaning up memory type warnings

### DIFF
--- a/hdf/src/dfp.c
+++ b/hdf/src/dfp.c
@@ -35,7 +35,7 @@ static uint16 Writeref = 0;
 static uint16 Refset   = 0; /* Ref of palette to get next */
 static uint16 Lastref  = 0; /* Last ref read/written */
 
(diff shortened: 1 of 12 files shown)


hdf/src/dfp.c
```diff
@@ -35,7 +35,7 @@ static uint16 Writeref = 0;
 static uint16 Refset   = 0; /* Ref of palette to get next */
 static uint16 Lastref  = 0; /* Last ref read/written */
 
-static char Lastfile[DF_MAXFNLEN] = ""; /* last file opened */
+static char Lastfile[DF_MAXFNLEN + 1] = ""; /* last file opened */
```
derobins, line 38: Same off-by-1 problem as in the previous warning cleanup work

derobins:
FWIW, mfgr.c is now almost warning-free and was previously one of the worst offenders in hdf/src

byrnHDF (approved):

PR #713, opened 2024-04-01
Bring over HDF5 warning macros + cleanup

derobins:
* Brings over H5_GCC_CLANG_DIAG_OFF, etc. macros
* Clean up more warnings
* Suppress a few unavoidable warnings, mainly functions taking const pointers via in/out pointers when the context is "write" and the buffer will be unmodified. We can't fix this without public API changes.


derobins:
Fewer than 900 warnings with this PR:

```
Total unique [non-ignored] warnings: 896                                         
Ignored notes / supplemental warning lines [not counted in unique warnings]: 0   
Duplicated warning lines [not counted in unique warnings]: 79                    
Total ignored warnings: 0                                                        
Total unique kinds of warnings: 29                                               
Total files with warnings: 89  
```

byrnHDF (approved):

PR #716, opened 2024-04-04
Fix build failure when szip is enabled.

sebastic:
The Debian package build for 4.3.20 failed due not linking libsz:
```
/usr/bin/ld: ../../hdf/src/.libs/libdf.so: undefined reference to `SZ_encoder_enabled'
/usr/bin/ld: ../../hdf/src/.libs/libdf.so: undefined reference to `SZ_BufftoBuffCompress'
/usr/bin/ld: ../../hdf/src/.libs/libdf.so: undefined reference to `SZ_BufftoBuffDecompress'
```
```
/usr/bin/ld: hrepack_parse.o: undefined reference to symbol 'SZ_encoder_enabled'
/usr/bin/ld: /lib/x86_64-linux-gnu/libsz.so.2: error adding symbols: DSO missing from command line
```
```
/usr/bin/ld: hrepacktst.o: undefined reference to symbol 'SZ_encoder_enabled'
/usr/bin/ld: /lib/x86_64-linux-gnu/libsz.so.2: error adding symbols: DSO missing from command line
```
The changes in this PR resolve this, but I'm not convinced it's the best solution as more than `-lsz` is included in `$LIBS`.

You may want to commit a better fix.

### DIFF
--- a/hdf/src/Makefile.am
+++ b/hdf/src/Makefile.am
@@ -34,6 +34,8 @@ libdf_la_SOURCES = $(CSOURCES)
 include_HEADERS = $(CHEADERS)
 endif
 
+libdf_la_LIBADD = @LIBS@
+
 # The following is a workaround. Since Fortran is included in this 
 # Makefile.am, automake will always try to use the Fortran linker, even when
 # fortran has been disabled in configure. The Fortran linker gets confused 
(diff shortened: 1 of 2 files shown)


hdf/src/Makefile.am
```diff
@@ -34,6 +34,8 @@ libdf_la_SOURCES = $(CSOURCES)
 include_HEADERS = $(CHEADERS)
 endif
 
+libdf_la_LIBADD = @LIBS@
```
byrnHDF, line 37: Where does @LIBS@ come from?

byrnHDF: Okay I found it used in the mfhdf/fortran Makefile.am - so OKAY

byrnHDF (approved):

PR #723, opened 2024-06-06
CMake: refactor for clarity and generality

scivision:
Each commit is independent, combined into one PR for convenience

* add_definitions() => add_compile_definitions() recommended for generality across compilers
* set policy implicitly, remove no-op policies already implied
* remove redundant directory checks


byrnHDF (approved):
The compile definitions need to move to targets eventually.

PR #772, opened 2024-10-05
Remove the NetCDF 2.3.2 API

derobins:
From both Fortran and C

See the RELEASE.txt note for a full description of the changes.


bmribler:
I would get inputs from Kent and Joe, and maybe Aleksandar on this.

derobins:
> I would get inputs from Kent and Joe, and maybe Aleksandar on this.

I've asked everyone I can think of, and nobody uses the NetCDF API we export and I've been announcing this for years. It's also not even documented. Anyone who is using our NetCDF interface can always use 4.3.0.

bmribler (approved):

byrnHDF (approved):

PR #781, opened 2024-10-07
Revert "Remove HDF-EOS2 CI action"

hyoklee:
Reverts HDFGroup/hdf4#751

### DIFF
--- a/.github/workflows/hdfeos2.yml
+++ b/.github/workflows/hdfeos2.yml
@@ -1,0 +1,40 @@
+name: hdfeos
+
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+    branches: [ master ]
+    paths-ignore:
+    - '.github/CODEOWNERS'
+    - 'doc/**'
+    - 'release_docs/**'
+    - 'COPYING'
+    - '**.md'
+
+jobs:
+  build:
+    name: Build hdfeos
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


schwehr:
It would be good to explain in detail why that removal should now be reverted.

byrnHDF (approved):

derobins:
> It would be good to explain in detail why that removal should now be reverted.

It was only removed because HDF-EOS2/5 went behind a login wall, causing the tests to fail when we couldn't get the code.

derobins (changes requested):
Why is this test failing? Why is the HDF-EOS2 compile barfing on missing C standard library calls?

lrknox:
> Why is this test failing? Why is the HDF-EOS2 compile barfing on missing C standard library calls?

The (first) error:
SetupSwath.c:97:38: error: ‘NC_UNLIMITED’ undeclared (first use in this function); did you mean ‘SD_UNLIMITED’?
   97 |     status = SWdefdim(SWid, "Unlim", NC_UNLIMITED);
   
NC_UNLIMITED was previously defined here:
    mfhdf/src/netcdf.h.in:#define NC_UNLIMITED 0L
    
File no longer exists - probably removed with netcdf 2.3.2 removal.

HDFEOS2 can't build with current HDF4 master.  Current hdfeos2 code will require HDF4.3.0 or earlier.

PR #796, opened 2025-01-06
Update CMake and libaec to follow hdf5 conventions

byrnHDF:


derobins:
Does this need a note in RELEASE.txt concerning the change to DHDF4_USE_SANITIZER?

byrnHDF:
> Does this need a note in RELEASE.txt concerning the change to DHDF4_USE_SANITIZER?

No because this was changed previously, but not all the references were updated.

derobins (approved):

PR #798, opened 2025-01-15
Remove autotools build system

byrnHDF:


.gitignore
```diff
@@ -10,25 +10,6 @@
...
-bin/missing
-bin/test-driver
-/configure
-hdf/src/h4config.h.in
-m4/libtool.m4
-m4/ltoptions.m4
-m4/ltsugar.m4
-m4/ltversion.m4
-m4/lt~obsolete.m4
-
```
derobins, line 31: autom4te.cache (below) is also an Autotools file, but we can remove that later

.gitattributes
derobins, line 1: There are also Makefile.am/in lines that should be removed

.github/workflows/main-cmake.yml
derobins, line 1: After we do the delete, all the -cmake prefixes could be dropped from the names

.gitignore
derobins, line 1: Need to also remove the Makefile.in line

release_docs/RELEASE.txt
```diff
@@ -38,6 +38,8 @@ New features and changes
 ========================
     Configuration:
     -------------
+    - Removed support for autotools build system.
```
derobins, line 41: Might also want to have another line where we mention that bin/cmakehdf4 was removed. That probably should have been done in a separate PR since it's not really an Autotools thing.

derobins:
There's a couple of comments where we should do a final tidy before we merge the PR. Should only take a few minutes.

byrnHDF:
Fixed the Makefile and gitattributes

derobins (approved):

Issue #18, opened 2022-03-28
`libdf` missing in CMake build


sbrodehl:
Hi there!

It looks like the CMake build is missing `libdf`, which I was able to build with autogen.
Is that possible?

Thanks!

sbrodehl:
Short follow up question: Has `libdf` been renamed to `libhdf` in the CMake build?

byrnHDF:
Yes, ever since CMake was added, I think.

sbrodehl:
Thanks for getting back to me.
Is that something which needs fixing?

byrnHDF:
CMake could create a link.

derobins:
Will a link work on Windows? And yes, we should create a link.

byrnHDF:
Closed by #801

PR #801, opened 2025-01-28
Fix issue #18 by creating a link to libdf

byrnHDF:


bmribler:
So LIB vs LIBSH is what determines hdf or mfhdf libraries?

byrnHDF:
> So, LIB vs LIBSH is what determines hdf or mfhdf libraries?

no, static vs shared

bmribler:
Oh, ok, so which line is the fix for this issue?

byrnHDF:
> Oh, ok, so which line is the fix for this issue?

Not sure what you need, the changes are the minimal needed so that libdf is in the installtion

bmribler (approved):

PR #813, opened 2025-04-28
Fix hdp to display special variables correctly

bmribler:
hdp used SDcheckempty to detect empty datasets. However, since netCDF variables don't have reference number, hdp needs to use the size of the record variable (i.e., unlimited dimension) to determine whether the variable is empty or not.  This fixed HDFFR-1293.

This PR also fixed hdp to display variables with rank = 0 properly, part of HDFFR-1291.

For internal use only:
HDFFR-1293
HDFFR-1291 partially


mfhdf/dumper/CMakeTests.cmake
```diff
@@ -249,6 +253,13 @@ ADD_H4_TEST (dumpsds-17 0 dumpsds -k -h -i 39,36 -n data34,data27 -r 36,37 -i 0,
 # Test 18 reads a few small datasets in a netCDF file
 ADD_H4_TEST (dumpsds-18 0 dumpsds -i 0,1,2 Roy.nc)
 
+# Test 19 displays a message when the file is a netCDF 64-bit file
+ADD_H4_TEST (dumpsds-19 0 dumpsds Roy-64.nc)
```
bmribler, line 257: This test was never added to CMake build

byrnHDF (approved):

Issue #825, opened 2025-06-06
Memory leak in NC_reset_maxopenfiles


bmribler:
==687765== 8,168 bytes in 1 blocks are still reachable in loss record 35 of 35
==687765== at 0x4848899: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==687765== by 0x148CF4: NC_reset_maxopenfiles (file.c:137)
==687765== by 0x149143: NC_open (file.c:265)
==687765== by 0x1492A5: nccreate (file.c:304)
==687765== by 0x14F363: SDstart (mfsd.c:319)
==687765== by 0x129238: test_max_open_files (tfile.c:260)
==687765== by 0x129B7E: test_files (tfile.c:442)
==687765== by 0x111C43: main (hdftest.c:1310)

PR #826, opened 2025-06-06
Fix a memory leak

bmribler:
- An if statement used the water mark variable instead of the number of opened files to determine when to release the list of file structs.  As a result, sometime the list was never released.  Fixed to use the correct variable.
- After that fix, some tests produced segfault because there was an extra call to release the list of file structs.  The extra code was removed.
- Other code tidying

This PR addressed GH-825


bmribler:
Only a few real changes, many were just style/format.

byrnHDF (approved):

Issue #818, opened 2025-05-09
hdf/test/testhdf produced Valgrind errors


bmribler:
==686741== Memcheck, a memory error detector
==686741== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==686741== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==686741== Command: ./testhdf
==686741==

For help use: testhdf -help
Built with HDF Library Version: 4.3.1, HDF Version 4.3 Release 1-1, April 11, 2025

Testing  -- Bit-Vector routines (bitvect)
Testing  -- Threaded Balanced Binary Trees (tbbt)
Testing  -- VERSION OF LIBRARY (vers)
Testing  -- HFILE (hfile)
Testing  -- HFILE LIMITS (hfile1)
Testing  -- HBLOCKS (hblocks)
Testing  -- EXTERNAL ELEMENTS (extelt)
Testing  -- COMPRESSED ELEMENTS (comp)
Testing  -- Chunks (chunks)
Testing  -- BIT I/O (bitio)
Testing  -- 8BIT RASTER IMAGE INTERFACE (8bit)
==686741== Syscall param write(buf) points to uninitialised byte(s)
==686741==    at 0x4A1C887: write (write.c:26)
==686741==    by 0x4992EEC: _IO_file_write@@GLIBC_2.2.5 (fileops.c:1180)
==686741==    by 0x49949E0: new_do_write (fileops.c:448)
==686741==    by 0x49949E0: _IO_new_do_write (fileops.c:425)
==686741==    by 0x49949E0: _IO_do_write@@GLIBC_2.2.5 (fileops.c:422)
==686741==    by 0x4995764: _IO_switch_to_get_mode (genops.c:166)
==686741==    by 0x4992942: _IO_file_seekoff@@GLIBC_2.2.5 (fileops.c:926)
==686741==    by 0x498F942: fseek (fseek.c:36)
==686741==    by 0x1F6B58: HPseek (hfile.c:3240)
==686741==    by 0x1F84E2: HTPsync (hfiledd.c:426)
==686741==    by 0x1F37BC: HIsync (hfile.c:1837)
==686741==    by 0x1EE879: Hclose (hfile.c:448)
==686741==    by 0x1C3C2D: DFR8Iputimage (dfr8.c:555)
==686741==    by 0x1C3D7E: DFR8addimage (dfr8.c:637)
[...]

PR #831, opened 2025-06-22
Fix uninitialized memory in test

bmribler:
GH-818 specified two memory issues

- Conditional jump or move depends on uninitialised value(s)
    This one appears to be caused by the JPEG library optimization.  The buffers
    in the related HDF4 test and library code were properly initialized.  A workaround to
    get rid of the valgrind error is using the environment variable JSIMD_FORCENONE.
    It is documented in a RELEASE.txt entry.
- Syscall param write(buf) points to uninitialised byte(s)
    Initializing a buffer fixed this error.            

This PR also makes a few other improvements with variable initialization.

Addresses GH-818

### DIFF
--- a/hdf/src/dfr8.c
+++ b/hdf/src/dfr8.c
@@ -44,7 +44,7 @@
 #include "dfrig_priv.h"
 
 /* Private constants */
-#define PALETTE_SIZE    768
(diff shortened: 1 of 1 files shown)


bmribler:
RELEASE.txt entry is coming...

PR #868, opened 2026-03-25
Fixed SDgetcompinfo incorrect behavior

bmribler:
When SZIP library is not available, SDgetcompinfo() fails even though it only needs to retrieve compression metadata, not actually compress/decompress data. The function should succeed in returning compression information regardless of whether the SZIP library is present.

This PR removes check from HCIinit_coder() to allow metadata queries without the szip library.
Fixes #867 

Known issue: Reading szip-compressed data from files opened with DFACC_RDWR when only decoder is available may fail. This is a pre-existing architectural issue where file open mode is conflated with operation type. This issue turned out to be too complicated to include in this PR, it will be addressed in a future fix, GH #870.


bmribler:
PR https://github.com/HDFGroup/hdf4/pull/891 replaces this PR for master

PR #879, opened 2026-06-07
Fix SZIP/libaec detection in CMakeFilters.cmake

bmribler:
The library was available but CMake couldn't find it because there were multiple minor mistakes.
- Add missing CONFIG keyword to find_package for libaec (the comment already said "Search pure Config mode" but the keyword was absent)
- Fix H4_SZIP_FOUND assignment to use libaec_FOUND directly instead of ${${LIBAEC_PACKAGE_NAME}_FOUND}, which was expanding incorrectly due to case mismatch
- Fix verbose message reporting SZIP_FOUND instead of H4_SZIP_FOUND


CMakeFilters.cmake
```diff
@@ -252,8 +252,8 @@ if (HDF4_ENABLE_SZIP_SUPPORT)
     set(libaec_USE_STATIC_LIBS ${HDF4_USE_LIBAEC_STATIC})
     set(SZIP_FOUND FALSE)
     # Search pure Config mode, there is not a FindSZIP module available
-    find_package (${LIBAEC_PACKAGE_NAME} NAMES ${LIBAEC_PACKAGE_NAME}${HDF_PACKAGE_EXT} OPTIONAL_COMPONENTS ${LIBAEC_SEARCH_TYPE})
-    set(H4_SZIP_FOUND ${${LIBAEC_PACKAGE_NAME}_FOUND})
+    find_package (${LIBAEC_PACKAGE_NAME} NAMES ${LIBAEC_PACKAGE_NAME}${HDF_PACKAGE_EXT} CONFIG OPTIONAL_COMPONENTS ${LIBAEC_SEARCH_TYPE})
```
jhendersonHDF, line 255: This is easy to miss, but from the CMake docs "The CONFIG option, the synonymous NO_MODULE option, or the use of options not specified in the [basic signature](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature) all enforce pure Config mode." In this case, using NAMES should be forcing config mode search. That said, I agree with adding CONFIG because its more obvious.

Also, in HDF5 there's now a module for finding libaec that can handle both module mode and config mode finds of libaec which so far has been more reliable in finding it. You can use that if you want, though some testing should be done.

CMakeFilters.cmake
```diff
@@ -252,8 +252,8 @@ if (HDF4_ENABLE_SZIP_SUPPORT)
     set(libaec_USE_STATIC_LIBS ${HDF4_USE_LIBAEC_STATIC})
     set(SZIP_FOUND FALSE)
     # Search pure Config mode, there is not a FindSZIP module available
-    find_package (${LIBAEC_PACKAGE_NAME} NAMES ${LIBAEC_PACKAGE_NAME}${HDF_PACKAGE_EXT} OPTIONAL_COMPONENTS ${LIBAEC_SEARCH_TYPE})
-    set(H4_SZIP_FOUND ${${LIBAEC_PACKAGE_NAME}_FOUND})
```
jhendersonHDF, line 256: This old variable is still the correct one to be used here. The package name can be changed by the configurer, which would break with these changes. The first argument to find_package is the "namespace" for the variables set.

jhendersonHDF (changes requested):
The logic as currently written should be correct. My best guess is that HDF4 is failing to find libaec for similar reasons that made https://github.com/HDFGroup/hdf5/blob/develop/config/cmake/Findlibaec.cmake necessary. I think mirroring what HDF5 is currently doing would work best. See https://github.com/HDFGroup/hdf5/pull/6152 for reference.

CMakeFilters.cmake
bmribler: OK.  Just so you know, I kept getting error that libaec wasn't found until I made the changes.

jhendersonHDF: It's likely there was something else in the build environment that was off. `LIBAEC_PACKAGE_NAME` expands to "libaec" by default, so `${${LIBAEC_PACKAGE_NAME}_FOUND}` and `${libaec_FOUND}` will expand to the same thing. Doing a quick build I can see that CMake is already performing a CONFIG-only search, which means you can point it to libaec by setting `libaec_DIR` to the directory containing `libaec-config.cmake` if libaec was built with CMake and CMake can't find it, but this is usually inconvenient. If libaec wasn't built with CMake, the current logic in HDF4 won't be able to find it outside of hacks. That's where the Findlibaec.cmake module helps.

bmribler: I built libaec with CMake and used libaec_DIR.  However, now I start to wonder whether trying multiple combinations gave me the wrong conclusion.  I'll start from scratch with those two in place first and see how it does.

jhendersonHDF: For libaec installed in system directories, setting libaec_DIR shouldn't usually be necessary. It's mostly for libaec installed in places that CMake isn't searching in, either by default or paths specified by the configurer.

bmribler:
Verified that the issue was user-error.

PR #881, opened 2026-06-10
Add new APIs to address vulnerabilities in the V API

### LINKED ISSUE #872: Stack buffer overflow in Vgetname, Vgetclass, and Vinquire via unbounded strcpy (CWE-121/CWE-787)
Description
Three functions in hdf/src/vgp.c perform unbounded strcpy into
caller-supplied buffers with no length parameter:

Vgetname() — line 828: strcpy(vgname, vg->vgname)
Vgetclass() — similar pattern in Vgetclass()
Vinquire() — strcpy(vgname, vg->vgname) with no NULL check
on vg->vgname, adding a null pointer dereference on top
vg->vgname is heap-allocated from a uint16 length field read
directly from the HDF file, meaning a malicious HDF file can
supply a name up to 65535 bytes long.

A vulnerable caller in HDF-EOS2 (SWfinfo in SWapi.c) passes a
fixed 80-byte stack buffer, which a crafted file can overflow.

Suggested fix: add a maxlen parameter to all three functions and
replace strcpy with strncpy + explicit null terminator.

bmribler:
The existing Vgroup APIs Vgetname, Vgetclass, and Vinquire posed multiple vulnerabilities due to the lack of size checks.  Because changing their signature will break compatibilities, three new APIs are added in place of their deprecation instead.  New APIs are:

int Vgetname40(int32 vkey, char *vgname, size_t *buf_size);
int Vgetclass40(int32 vkey, char *vgclass, size_t *buf_size);
int Vinquire40(int32 vkey, int32 *nentries, char *vgname, size_t *buf_size);

Fixes GH #872

The removed APIs are no longer needed with the new argument buf_size in the new APIs.
int32 Vgetnamelen(int32 vkey, uint16 *name_len); /* deprecated in 4.0 */
int32 Vgetclassnamelen(int32 vkey, uint16 *classname_len); /* deprecated in 4.0 */

### DIFF
--- a/hdf/src/hproto.h
+++ b/hdf/src/hproto.h
@@ -1402,15 +1402,21 @@ HDFLIBAPI int32 Vgetid(HFILEID f, int32 vgid);
 
 HDFLIBAPI int32 Vgetnext(int32 vkey, int32 id);
 
-HDFLIBAPI int32 Vgetname(int32 vkey, char *vgname);
+HDFLIBAPI int Vgetname40(int32 vkey, char *vgname, size_t *buf_size); /* in 4.0 */
 
-HDFLIBAPI int32 Vgetnamelen(int32 vkey, uint16 *name_len);
+HDFLIBAPI int Vgetclass40(int32 vkey, char *vgclass, size_t *buf_size); /* in 4.0 */
 
-HDFLIBAPI int32 Vgetclassnamelen(int32 vkey, uint16 *classname_len);
+HDFLIBAPI int Vinquire40(int32 vkey, int32 *nentries, char *vgname, size_t *buf_size); /* in 4.0 */
 
-HDFLIBAPI int32 Vgetclass(int32 vkey, char *vgclass);
+HDFLIBAPI int32 Vgetname(int32 vkey, char *vgname); /* deprecated in 4.0 */
 
-HDFLIBAPI int Vinquire(int32 vkey, int32 *nentries, char *vgname);
+HDFLIBAPI int32 Vgetnamelen(int32 vkey, uint16 *name_len); /* deprecated in 4.0 */
+
+HDFLIBAPI int32 Vgetclassnamelen(int32 vkey, uint16 *classname_len); /* deprecated in 4.0 */
+
+HDFLIBAPI int32 Vgetclass(int32 vkey, char *vgclass); /* deprecated in 4.0 */
+
+HDFLIBAPI int Vinquire(int32 vkey, int32 *nentries, char *vgname); /* deprecated in 4.0 */
 
 HDFLIBAPI int32 Vdelete(int32 f, int32 ref);
 
@@ -1440,6 +1446,17 @@ HDFLIBAPI int Vdeletetagref(int32 vkey, /* IN: vgroup key */
                             int32 ref /* IN: ref to delete in vgroup */);
 
 HDFLIBAPI int VPshutdown(void);
+
+/* These are deprecated in version 4.0 */
+HDFLIBAPI int32 Vgetname(int32 vkey, char *vgname);
+
+HDFLIBAPI int32 Vgetnamelen(int32 vkey, uint16 *name_len);
+
+HDFLIBAPI int32 Vgetclassnamelen(int32 vkey, uint16 *classname_len);
+
+HDFLIBAPI int32 Vgetclass(int32 vkey, char *vgclass);
+
+HDFLIBAPI int Vinquire(int32 vkey, int32 *nentries, char *vgname);
 
 /*
  ** from vparse.c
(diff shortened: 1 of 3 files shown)


hdf/src/hproto.h
```diff
@@ -1402,15 +1402,21 @@ HDFLIBAPI int32 Vgetid(HFILEID f, int32 vgid);
 
 HDFLIBAPI int32 Vgetnext(int32 vkey, int32 id);
 
-HDFLIBAPI int32 Vgetname(int32 vkey, char *vgname);
+HDFLIBAPI int Vgetname40(int32 vkey, char *vgname, size_t *buf_size); /* in 4.0 */
```
schwehr, line 1405: Why are there  `40` added to these function names? Maybe something in the name that implies it knows the string length?

hdf/src/hproto.h
```diff
@@ -1402,15 +1402,21 @@ HDFLIBAPI int32 Vgetid(HFILEID f, int32 vgid);
...
-HDFLIBAPI int32 Vgetclassnamelen(int32 vkey, uint16 *classname_len);
+HDFLIBAPI int Vinquire40(int32 vkey, int32 *nentries, char *vgname, size_t *buf_size); /* in 4.0 */
 
-HDFLIBAPI int32 Vgetclass(int32 vkey, char *vgclass);
+HDFLIBAPI int32 Vgetname(int32 vkey, char *vgname); /* deprecated in 4.0 */
 
-HDFLIBAPI int Vinquire(int32 vkey, int32 *nentries, char *vgname);
+HDFLIBAPI int32 Vgetnamelen(int32 vkey, uint16 *name_len); /* deprecated in 4.0 */
+
+HDFLIBAPI int32 Vgetclassnamelen(int32 vkey, uint16 *classname_len); /* deprecated in 4.0 */
```
schwehr: Why is this also down below?

schwehr: Consider requiring the user to `#define` something to allow use of deprecated calls and default it to set in the CMake system?

hdf/src/vgp.c
```diff
@@ -3169,3 +3168,237 @@ Vgetvgroups(int32    id,       /* IN: file id or vgroup id */
...
+
+    /* Clear error stack */
+    HEclear();
+
+    /* Check arguments */
+    if (HAatom_group(vkey) != VGIDGROUP || buf_size == NULL)
+        HGOTO_ERROR(DFE_ARGS, FAIL);
+
+    /* Get the vgroup struct for access */
+    if ((vg = VIGet_vgdesc(vkey)) == NULL)
```
schwehr, line 3244: Switching style how it checks against `NULL`. Above was `if (NULL == `. Use the same style across all.

hdf/src/vgp.c
```diff
@@ -3169,3 +3168,237 @@ Vgetvgroups(int32    id,       /* IN: file id or vgroup id */
...
+
+    /* Get the number of entries in vgroup, if requested */
+    if (nentries != NULL)
+        *nentries = (int32)vg->nvelt;
+
+    /* If buf_size is provided, handle name copy or length query */
+    if (buf_size != NULL) {
+        if (vgname == NULL || *buf_size == 0) {
+            *buf_size = name_len; /* return the name length */
+        }
```
schwehr, line 3389: Maybe return here so you don't have to indent for the else?

schwehr (changes requested):
Some quick thoughts on the changes.

Edit: Thank you! These kinds of improvements are key for the long term stability of this old, but important, code base.

hdf/src/hproto.h
bmribler: My intention for 40 is they are introduced in 4.0. :D

hdf/src/hproto.h
bmribler: We don't have plan to do much more in HDF4.  4.0 is very likely to be the last release.  Applications can continue using 3.x if the user chooses to use deprecated APIs.  Please see the [note about deprecating](https://www.hdfgroup.org/solutions/hdf4/).

bmribler: I removed them. Thanks!!

hdf/src/vgp.c
bmribler: I prefer (... == NULL) but I'll use the existing style for now.

hdf/src/hproto.h
bmribler: Actually, I'm reconsidering this position...

bmribler: We'll have a clean break, no deprecated functions, cleaner API for 4.0

hdf/src/vgp.c
schwehr: Sounds good. I suggest a follow up PR to switch to `== NULL`  as I agree with your take on what is better these days.

bmribler:
@schwehr Thank you for reviewing this PR!  I'll close it but create another PR soon, and apply your comments there, where appropriate.

PR #890, opened 2026-08-07
Code and test cleanup

bmribler:
#### Purpose
Resource handling for VSattach/Vstart, improvement of JPEG tests, and a valgrind false-positive note.

#### Description
- Adds missing VSdetach/Vend calls for corresponding VSattach/Vstart calls in source and test code, fixing memory leaks/unreachable blocks flagged by valgrind.
- Replaces static JPEG reference-data comparisons in mgr.c (jpeg_8bit_j80, jpeg_24bit_j80) with runtime comparisons against data decoded independently via decomp_using_jpeglib(), fixing test failures caused by decoding differences between JPEG 6b (used to generate the original reference data) and newer JPEG libraries. All three images in each test file are now verified, not just the first.
- Adds an in-code note in the mgr test documenting a known valgrind false positive ("Conditional jump or move depends on uninitialised value(s)") caused by Valgrind's limited support for libjpeg-turbo's SIMD code — confirmed benign by rerunning with JSIMD_FORCENONE=1.

#### Testing
Valgrind output is now clean across the testsuite (aside from the documented, confirmed-benign SIMD false positive).


hdf/test/mgr.c
```diff
@@ -3791,15 +3525,22 @@ test_mgr_old_c(int flag)
...
-        int32 start[2];                 /* start of image data to grab */
-        int32 stride[2];                /* stride of image data to grab */
+        int32 offset = 0, length = 0;
+        int32 n_entries = 0;
+        uint8 jpeglib_readbuf[JPEGY * JPEGX];
+        /* buffer for data read by JPEG function */
+        int32 start[2];  /* start of image data to grab */
+        int32 stride[2]; /* stride of image data to grab */
 
         /* Get the first image in this file */
```
lrknox, line 3542: With the changes, is it reading multiple images, not just the first?  Should the comment read "Get the next image..."?

hdf/test/mgr.c
```diff
@@ -3808,11 +3549,20 @@ test_mgr_old_c(int flag)
...
+        /* Compare data decompressed by HDF against that by JPEG lib, the buffers
+           should be identical */
+        if (memcmp(image, jpeglib_readbuf, JPEGY * JPEGX)) {
+            fprintf(stderr, "Error reading data for 8-bit JPEG compressed image %d\n", idx);
+            print_mismatched(image, jpeglib_readbuf, JPEGY * JPEGX);
             num_errs++;
-        } /* end if */
+        }
 
         /* Close the empty image */
```
lrknox, line 3567: It's just a comment, but "empty" is probably inaccurate.  I suppose it could just be removed.

hdf/test/mgr.c
```diff
@@ -3866,12 +3625,20 @@ test_mgr_old_e(int flag)
...
+           should be identical */
+        if (memcmp(image, jpeglib_readbuf, JPEGY * JPEGX * 3)) {
+            fprintf(stderr, "Error reading data for 24-bit JPEG compressed image #%d\n", idx);
+            print_mismatched(image, jpeglib_readbuf, JPEGY * JPEGX * 3);
             num_errs++;
- */
-        } /* end if */
+        }
 
         /* Close the empty image */
```
lrknox: It's just a comment, but "empty" is probably inaccurate.  I suppose it could just be removed.

mfhdf/test/tmixed_apis.c
```diff
@@ -209,8 +209,19 @@ test_nonSDAPI_ids()
...
+        GRendaccess(ri_id);
+    if (gr_id != FAIL)
+        GRend(gr_id);
+    if (fid != FAIL) {
+        Vend(fid);
+        Hclose(fid);
+    }
+
+    /* Return the number of errors that's been kept track of so far */
     return num_errs;
```
lrknox, line 225: I only see num_errs set to 0 in line 132 and returned in line 225.  Is it ever changed from 0?

bmribler: It's incremented by the CHECK, VERIFY, etc... macros.

lrknox (approved):

PR #893, opened 2026-08-13
Do not free the memory being used for the frontier global in dfimcompc init().

schwehr:


fortnern:
Why does this not cause segfaults in normal use without the fix?

schwehr:
> Why does this not cause segfaults in normal use without the fix?

I am don't know why. I haven't tried to check coverage.

bmribler:
@schwehr I'm surprised to see that IMCOMP code is still touched. I was told there were data still out there but hadn't heard it from users.

schwehr:
> @schwehr I'm surprised to see that IMCOMP code is still touched. I was told there were data still out there but hadn't heard it from users.

I have not clue. This was a purely by inspection, I've never heard of IMCOMP that I can remember before today.

> IMCOMP is an early, specialized raster image compression method used primarily for compressing 8-bit raster images (RIS8) along with an associated color palette.

fortnern (approved):
Approved, though if the code is no longer used maybe it should be removed instead? This is fine for now.

PR #894, opened 2026-08-14
Fix to prevent resource leaks in hdf tests

bmribler:
- Implement a unified single-point-of-exit pattern across all test functions to prevent resource leaks.

    * Add 'goto done;' to CHECK, VERIFY, and RESULT macros to jump directly to cleanup on failure.
    * Add a 'done:' label to every test function using CHECK, VERIFY, or RESULT to safely close opened objects and files.
    * Remove the now-obsolete CHECK_VOID/VERIFY_VOID macros, no longer needed now that CHECK/VERIFY use goto done uniformly.
    * Introduce ENDACCESS/ENDANNO/ENDAN/ENDRI/ENDGR macros mirroring mfhdf/test's ENDSDS/ENDSD pattern.

- Audit every touched function for memory/resource leaks now reachable via goto done, and fix pre-existing bugs such as leaked buffers on failures and a double-close bug.

- Replace local error-counting variables with direct num_errs updates, which had hidden some test failures on early exit.

Note: chunks.c currently has more than 2000 lines in one function. It is planned to be split into smaller functions.


mfhdf/test/hdftest.h
```diff
@@ -73,45 +73,45 @@
...
    error message */
-#define VERIFY_CHAR(item, value, test_name)                                                                  \
+#define VERIFY_CHAR(val, x, test_name)                                                                       \
     do {                                                                                                     \
-        if (strcmp(item, value) != 0) {                                                                      \
-            fprintf(stderr, "*** UNEXPECTED VALUE from %s is <%s> at line %4d in %s\n", test_name, item,     \
-                    (int)__LINE__, __FILE__);                                                                \
+        if (strncmp(x, val, strlen(x)) != 0) {                                                               \
+            fprintf(stderr, "*** UNEXPECTED VALUE from %s is <%s> (expected <%s>) at line %4d in %s\n",      \
+                    test_name, val, x, (int)__LINE__, __FILE__);                                             \
```
lrknox, line 114: Could strncmp(x, val, strlen(x)) find val to be the same as x if x matches the first part of val, but val has more characters after that?  Is another condition strlen(val) == strlen(x) also needed?

bmribler: Good catch! Switching to strncmp was the wrong call. I'll fix it.

hdf/test/tvattr.c
```diff
@@ -75,132 +69,121 @@ static int  write_vattrs(void);
...
     }
     return SUCCEED;
-} /* create_vset_stuff */
+}
 
-/* test attribute routines */
+/*  write_vattr() adds attributes to the vgroups, vdata and
+ *  the fields of the vdata. It then creates new vdata and vgroup,
+ *  and adds attrs to them.  read_vattr reads the attributes back,
+ *  and check their correctness.
```
lrknox:
```suggestion
 *  and checks their correctness.
```

lrknox (approved):

schwehr:
https://github.com/HDFGroup/hdf4/issues/899 has a request for a minor follow up fix to this PR. Thanks for doing so much cleanup!

bmribler:
Thank you, @schwehr

PR #895, opened 2026-08-14
Set buf to NULL after free in vconv vimakecompat.

schwehr:
I think `vimakecompat`  can be simplified more, but this is the smallest change possible to avoid a possible double free situation.

### DIFF
--- a/hdf/src/vconv.c
+++ b/hdf/src/vconv.c
@@ -222,6 +222,7 @@ vimakecompat(HFILEID f)
         ret = Hputelement(f, VGDESCTAG, ref, (uint8 *)buf, bsize);
         free(buf);
         buf = NULL;
+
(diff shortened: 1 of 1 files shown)


fortnern (approved):

hdf/src/vconv.c
```diff
@@ -221,6 +221,7 @@ vimakecompat(HFILEID f)
 
         ret = Hputelement(f, VGDESCTAG, ref, (uint8 *)buf, bsize);
         free(buf);
+        buf = NULL;
```
fortnern, line 224:
```suggestion
        buf = NULL;

```
We should add a blank line to preserve readability

schwehr:
@fortnern Thanks for the touchup on the PR!

PR #901, opened 2026-08-23
Fix stack buffer overflow in Vgetname/Vgetclass/Vinquire

### LINKED ISSUE #872: Stack buffer overflow in Vgetname, Vgetclass, and Vinquire via unbounded strcpy (CWE-121/CWE-787)

bmribler:
Vgetname, Vgetclass, and Vinquire performed unbounded strcpy into caller-supplied buffers with no length parameter, allowing stack buffer overflows on names/classes longer than the caller anticipated.

Each function now takes a buf_size parameter and returns the actual name/class length on success or FAIL on error, using ptrdiff_t (instead of ssize_t) as the return type for Windows portability without a POSIX-type workaround. This makes Vgetnamelen and Vgetclassnamelen redundant, so they're removed.

Two utility functions were added, vgetvgname/vgetvgclass.  They return the name/class of a vgroup, allocating the buffer internally so the caller does not need to know the length of the name/class in advance.  Strictly intended for use within the library and tools only by including the private header file vg_priv.h; the caller is responsible for freeing the returned buffer.

BREAKING: callers must update call sites to pass buf_size to any of these three functions and replace the calls to Vgetnamelen/Vgetclassnamelen by the appropriate Vgetname/Vgetclass calls, passing in 0 and NULL for buffer size and buffer.

Fixes #872

### DIFF
--- a/HDF4Examples/C/VG/h4ex_VG_get_vgroup_info.c
+++ b/HDF4Examples/C/VG/h4ex_VG_get_vgroup_info.c
@@ -65,25 +65,25 @@ main()
              * moving to the next.
              */
             vgroup_id = Vattach(file_id, ref_array[lone_vg_number], "r");
-            if (Vgetnamelen(vgroup_id, &name_len) == FAIL)
-                printf("*** ERROR from Vgetnamelen\n");
+            if ((name_len = Vgetname(vgroup_id, 0, NULL)) == FAIL)
+                printf("*** ERROR from Vgetname to get length of vgroup name\n");
             vgroup_name = (char *)malloc(sizeof(char *) * (name_len + 1));
             if (vgroup_name == NULL) {
                 fprintf(stderr, "Not enough memory for vgroup_name!\n");
                 exit(1);
             }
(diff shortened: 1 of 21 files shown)


bmribler:
Redo by #903

PR #924, opened 2026-09-05
Fix pkg-config {include,lib}dir paths.

sebastic:
The `includedir` & `libdir` variables in the `pkg-config` file are incorrect when non-default paths are used using `HDF4_INSTALL_INCLUDE_DIR` & `HDF4_INSTALL_LIB_DIR`.

This is fixed by not hardcoding the subdirectories.

The resulting `hdf.pc` still has an issue, the private libs contain CMake targets:
```
Libs.private:  -lhdf -lmfhdf  -lm -l/usr/lib/x86_64-linux-gnu/libjpeg.so -lZLIB::ZLIB -llibaec::sz -llibaec::aec
```
I don't know how to fix this, so left this out of scope for this PR.
Hopefully someone more familiar with CMake knows how to resolve these targets.


jhendersonHDF:
https://github.com/HDFGroup/hdf5/pull/6220 may be of use if anyone wants to take this work on.

